PR #29, opened 2015-01-28
Print replicated and starting bundle counds. Print hints.

### LINKED ISSUE #28: Bundle Info Run vs Start
Bundle info should provide a `#Start` column in addition to `#Run` that distinguishes the number of bundles starting from the number that are running. At present `#Run` reports a bundle as running even when it is starting.

2m:
Fixes #28 and #26.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
@@ -105,11 +105,37 @@ object Column {
       bundles.map { bundle =>
         bundle.bundleInstallations.map { node =>
           val nodeIpAndPort = node.uniqueAddress.address.toString.dropWhile(_ != '@').drop(1)
-          if (bundle.bundleExecutions.map(_.host) contains node.uniqueAddress.address.host.get)
+          if (bundle.bundleExecutions.withFilter(_.isStarted).map(_.host) contains node.uniqueAddress.address.host.get)
             nodeIpAndPort.invert
           else
             nodeIpAndPort
         }
+      }
+  }
+
+  /**
+   * Displays the number of hosts where bundle is replicated.
+   */
+  case class Replicated(bundles: Seq[ConductRController.BundleInfo]) extends RegularColumn with RightJustified {
+    val title = "#REP"
+    val width = 7
+
+    val data: Seq[Seq[String]] =
+      bundles.map { bundle =>
+        List(bundle.bundleInstallations.size.toString)
+      }
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -105,7 +105,7 @@ object Column {
       bundles.map { bundle =>
         bundle.bundleInstallations.map { node =>
           val nodeIpAndPort = node.uniqueAddress.address.toString.dropWhile(_ != '@').drop(1)
-          if (bundle.bundleExecutions.map(_.host) contains node.uniqueAddress.address.host.get)
+          if (bundle.bundleExecutions.filter(_.isStarted).map(_.host) contains node.uniqueAddress.address.host.get)
```
2m: Yes, filter + map = collect, but this is more elegant IMO than:
```
import ....BundleExecution
...
... collect { case BundleExecution(host, isStarted) if isStarted => host }
```

hseeberger: But your solution iterates the collection twice!
Also, you can improve the function given to collect:
``` scala
collect { case BundleExecution(host, true) => host }
```

hseeberger: Alternatively you can use `withFilter` which avoids the double iteration.

2m: Yes. But only the values left out after the filter. :) Okay, I can change to collect.

2m: Ahh. Yes. With filter is nice here.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -105,7 +105,7 @@ object Column {
       bundles.map { bundle =>
         bundle.bundleInstallations.map { node =>
           val nodeIpAndPort = node.uniqueAddress.address.toString.dropWhile(_ != '@').drop(1)
-          if (bundle.bundleExecutions.map(_.host) contains node.uniqueAddress.address.host.get)
+          if (bundle.bundleExecutions.withFilter(_.isStarted).map(_.host) contains node.uniqueAddress.address.host.get)
```
huntc: Oh, how did @hseeberger let you get away with the infix notation? ;-)

hseeberger: Ouch, what a terrible mistake ;-)

PR #33, opened 2015-02-04
Handle bundle name

### LINKED ISSUE #32: Handle bundle names

2m:
Fixes #32.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/client/ConductRController.scala
+++ b/src/main/scala/com/typesafe/conductr/client/ConductRController.scala
@@ -25,6 +25,9 @@ import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 
 object ConductRController {
+
+  private[this] val BundlePath = "(.*?)(-[a-fA-F0-9]{0,64})?\\.zip".r
+
   /**
    * The Props for an actor that represents the ConductR's control endpoint.
    * @param address The address to reach the ConductR at.
@@ -94,17 +97,18 @@ object ConductRController {
     bundleDigest: String,
     configDigest: Option[String],
     bundleInstallations: Seq[BundleInstallation],
-    schedulingRequirement: SchedulingRequirement,
+    attributes: Attributes,
     bundleExecutions: Set[BundleExecution])
 
   /**
-   * Representation of bundle requirements to be scheduled.
+   * Representation of bundle attributes.
    */
-  case class SchedulingRequirement(
+  case class Attributes(
     nrOfCpus: Double,
     memory: Long,
     diskSpace: Long,
-    roles: Set[String])
+    roles: Set[String],
+    bundleName: String)
 
   /**
    * Descriptor of a node's bundle installation including its associated optional configuration.
@@ -143,6 +147,12 @@ object ConductRController {
         )
       )
     )
+
+  private def toBundleName(bundle: Uri): String = {
+    bundle.path.toString.reverse.takeWhile(_ != '/').reverse match {
+      case BundlePath(name, _) => name
+    }
+  }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


huntc:
Nice one! LGTM.

@hseeberger Would you be able to review also and then merge. After that would you mind releasing? Thanks.

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -143,6 +144,13 @@ object ConductRController {
         )
       )
     )
+
+  private def toBundleName(bundle: Uri): String = {
+    val BundlePath = "(.*?)(-[a-fA-F0-9]{0,64})?\\.zip".r
```
hseeberger: Instead of creating the regex every time this method is called, I'd rather move that to a val in the companion object.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -61,6 +61,18 @@ object Column {
 
     protected def space(length: Int): String =
       Stream.continually(" ").take(length).mkString
+
+    /**
+     * Add '... ' if string is over the allowed width.
+     */
+    private def ellipsize(s: String): String = {
+      val ellipsis = "... "
+      if (s.length >= width) {
```
hseeberger: Curly braces not needed.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -61,6 +61,18 @@ object Column {
...
+        s.take(width - ellipsis.length) + ellipsis
+      } else {
```
hseeberger: Same here.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -94,6 +106,17 @@ object Column {
   }
 
   /**
+   * Displays bundle name.
+   */
+  case class Name(bundles: Seq[ConductRController.BundleInfo]) extends RegularColumn {
+    val title = "NAME"
```
hseeberger: Move aS `final val Title = "NAME"` to the companion object.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
```diff
@@ -94,6 +106,17 @@ object Column {
...
+    val width = 30
```
hseeberger: Move as `final val Width = 30` to the companion object.

src/test/scala/com/typesafe/conductr/client/ConductRControllerSpec.scala
```diff
@@ -47,6 +47,21 @@ class ConductRControllerSpec extends WordSpec with Matchers with BeforeAndAfterA
...
 
+  "Bundle URI" should {
+    "be converted to bundle name" in {
+      val pathsToNames = Map(
+        "path/to/bundle.zip" -> "bundle",
+        "path/to/bundle-5ca1ab1e.zip" -> "bundle",
+        "path/to/bundle-1.0.0-M2.zip" -> "bundle-1.0.0-M2"
+      )
+
+      val toBundleName = PrivateMethod[String]('toBundleName)
```
hseeberger: Wow, that's awesome!!

src/test/scala/com/typesafe/conductr/client/ConductRControllerSpec.scala
```diff
@@ -47,6 +47,21 @@ class ConductRControllerSpec extends WordSpec with Matchers with BeforeAndAfterA
...
+      pathsToNames foreach {
+        case (path, name) => ConductRController.invokePrivate(toBundleName(Uri(s"file://$path"))) should be(name)
```
hseeberger: `shouldBe name`

hseeberger:
@2m, some minor changes needed. Please let me know once you have updated this PR, so I can release.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
2m: Both of these actually override a method in the parent.

2m:
Applied PR feedback. I did not move the `title` and `name` from each of the case classes because these fields override methods in the parent. So these abstract methods would still need to be implemented if these values would be moved to companion object as compile time constants.

src/main/scala/com/typesafe/conductr/sbt/console/Columns.scala
hseeberger: Ah, then please add an `override` modifier.

2m:
Added `overrides`.

hseeberger:
Released 0.18.0.

2m:
:+1:

PR #52, opened 2015-03-27
!makes APIs discoverable and introduces scripted tests

ktoso:
WIP - polishing this one up a bit still.


README.md
```diff
@@ -69,24 +69,24 @@ import NativePackagerKeys._
...
 _...otherwise you will get duplicate imports reported. This is because the new 1.0+ version uses sbt's auto plugin feature._
 
-Unless the ConductR is running at http://127.0.0.1:9005, and instead supposing that it is running at
-http://192.168.59.103:9005 then you will typically issue the following command from the sbt console:
+Unless the ConductR is running at `127.0.0.1:9005`, and instead supposing that it is running at
+`192.168.59.103:9005` then you will typically issue the following command from the sbt console:
 
 ```
-conductr http://192.168.59.103:9005
+conductr:set 192.168.59.103:9005
```
hseeberger: So `conductr` is a configuration now? Nice!
Yet I don't like `set`: Set what? Why not `control`?

ktoso: control sounds good :+1:

ktoso: or `control-server`

src/main/scala/com/typesafe/conductr/sbt/SbtTypesafeConductR.scala
```diff
@@ -247,12 +263,12 @@ object SbtTypesafeConductR extends AutoPlugin {
...
   // Actor system management and API
 
-  private val actorSystemAttrKey = AttributeKey[ActorSystem]("sbt-typesafe-conductr-actor-system")
+  private val actorSystemAttrKey = AttributeKey[ActorSystem]("sbt-typesafe-setConductrTask-actor-system")
 
   private def loadActorSystem(state: State): State =
     state.get(actorSystemAttrKey).fold {
       state.log.debug(s"Creating actor system and storing it under key [${actorSystemAttrKey.label}]")
-      val system = withActorSystemClassloader(ActorSystem("sbt-typesafe-conductr"))
+      val system = withActorSystemClassloader(ActorSystem("sbt-typesafe-setConductrTask"))
```
hseeberger: set-conductr-task, i.e. don't mix camel case and the other one (with hyphens).

version.sbt
```diff
@@ -1 +1,12 @@
-version in ThisBuild := "0.26.0-SNAPSHOT"
\ No newline at end of file
+import java.io.FileInputStream
+import java.util.Properties
+
+version in ThisBuild := {
+  val s = new FileInputStream(file("version.properties"))
```
hseeberger: Nice trick!

README.md
hseeberger: +1

src/main/scala/com/typesafe/conductr/sbt/SbtTypesafeConductR.scala
ktoso: right, thanks

ktoso:
refs https://github.com/sbt/sbt-typesafe-conductr/issues/50

README.md
```diff
@@ -28,7 +28,7 @@ A bundle will be produced from the native packager settings of this project. A b
 packager distribution and includes some component configuration. To load the bundle:
 
 ```bash
-loadBundle <HIT THE TAB KEY AND THEN RETURN>
+conductr:loadBundle <HIT THE TAB KEY AND THEN RETURN>
```
huntc: If we're going to change anything in terms of the structure of the commands then we should align more closely with the Python CLI - there's a ticket for this.

Thus the above command should be `conduct load`.

README.md
```diff
@@ -69,24 +69,24 @@ import NativePackagerKeys._
...
 _...otherwise you will get duplicate imports reported. This is because the new 1.0+ version uses sbt's auto plugin feature._
 
-Unless the ConductR is running at http://127.0.0.1:9005, and instead supposing that it is running at
-http://192.168.59.103:9005 then you will typically issue the following command from the sbt console:
+Unless the ConductR is running at `127.0.0.1:9005`, and instead supposing that it is running at
+`192.168.59.103:9005` then you will typically issue the following command from the sbt console:
 
 ```
-conductr http://192.168.59.103:9005
+conductr:controlServer 192.168.59.103:9005
```
huntc: We can abolish this command as a result of now having `CONDUCTR_IP` and `CONDUCTR_PORT`.

huntc:
@ktoso This PR has gone a little beyond what I had envisaged. I'm now hung up on the re-factoring of the commands. If we're going to do that then we need to align them with the Python CLI - I've received such feedback from others already. Can you please either remove the re-factoring of the commands or re-factor them such that they align with the Python CLI.

Thanks.

README.md
ktoso: I argue for `conductr:load`, which it is currently. The above it a typo in docs.

README.md
```diff
@@ -69,24 +69,24 @@ import NativePackagerKeys._
...
-Property     | Description
--------------|------------
-conductr     | Sets the ConductR's address to a provided url (defaults to env varialbe `CONDUCTR_IP:[CONDUCTR_PORT]` if set, otherwise uses the default: `http://127.0.0.1:9005`)
-loadBundle   | Loads a bundle and an optional configuration to the ConductR
-startBundle  | Starts a bundle given a bundle id with an optional absolute scale value
-stopBundle   | Stops all executions of a bundle given a bundle id
-unloadBundle | Unloads a bundle entirely (requires that the bundle has stopped executing everywhere)
+Property              | Description
+----------------------|------------
+conductr              | Sets the ConductR's address to a provided url (defaults to env varialbe `CONDUCTR_IP:[CONDUCTR_PORT]` if set, otherwise uses the default: `http://127.0.0.1:9005`)
```
ktoso: FIXME: this is controlServer

README.md
huntc: If the goal is discoverability, then do you agree that a `conduct` command achieves this goal given tab completion?

I'm trying to understand why having a scope would be better than a command, given that the latter more closely resembles our Python CLI.

ktoso: You're right that it's doable with commands, was not the immediate "sbt style" I would think of. 
It's more work but can be done if you think it's worth it

README.md
ktoso: related - CONDUCTR_*IP* seems wrong, isn't it "HOST" or "URL"? Can be a domain name, not just IP.

README.md
huntc: I think that having the command align with the Python CLI as much as reasonable makes sense, and was raised by someone else. I realise that it is more work, but I think it is worth it. Putting the effort aside, let's ask ourselves what the best approach would be. WDYT?

ktoso: As a scala dev I'd expect `namespace:...` to include both tasks and other keys I can `inspect` on, not only the DSL as seen in command line.

CLI-like commands would be an addition, these could use `conduct`, while keys are all in `conductr`.

huntc: With sbt-web we chose the approach that common commands were not namespaced in order to reduce typing.

When I think about it, the `conduct` command for the Python CLI is really about namespacing on the command line, so an sbt convention of namespacing all commands to align with the CLI sub commands sounds reasonable.

Let's just fix up the doc then and go with the namespaced approach then. :-)

huntc: @ktoso Any chance we can get this PR finished off? Thanks.

ktoso: yes sure, I do mean to pull through my outstanding things of course :)
on this one just the typo in docs (where this comments thread is attached to) as I understand the current consensus?

README.md
huntc: It must be an ip in other contexts eg for ConductR itself.

huntc:
Looking good with exception to your fixme

README.md
```diff
@@ -69,24 +69,23 @@ import NativePackagerKeys._
...
--------------|------------
-conductr     | Sets the ConductR's address to a provided url (defaults to env varialbe `CONDUCTR_IP:[CONDUCTR_PORT]` if set, otherwise uses the default: `http://127.0.0.1:9005`)
-loadBundle   | Loads a bundle and an optional configuration to the ConductR
-startBundle  | Starts a bundle given a bundle id with an optional absolute scale value
-stopBundle   | Stops all executions of a bundle given a bundle id
-unloadBundle | Unloads a bundle entirely (requires that the bundle has stopped executing everywhere)
+Property               | Description
+-----------------------|------------
+conductr:controlServer | Sets the ConductR's address to a provided url (defaults to env varialbe `CONDUCTR_IP:[CONDUCTR_PORT]` if set, otherwise uses the default: `http://127.0.0.1:9005`)
+conductr:loadBundle    | Loads a bundle and an optional configuration to the ConductR
```
huntc: @ktoso This looks as though it still needs changing. Any ETA? Thanks.

ktoso: ETA - now.
Rebased it as well.

PR #57, opened 2015-04-01
General tidying up given recent commits

huntc:


.gitignore
```diff
@@ -17,6 +17,7 @@ target
 
 # IntelliJ
 .idea
+.idea_modules
```
huntc: These suddenly appeared

build.sbt
```diff
@@ -31,4 +31,4 @@ repository in bintray := "sbt-plugins"
 bintrayOrganization in bintray := Some("sbt-typesafe-conductr")
 
 scriptedSettings
-scriptedLaunchOpts <+= version(v => s"-Dproject.version=$v")
+scriptedLaunchOpts <+= version apply { v => s"-Dproject.version=$v" }
```
huntc: Just being consistent with other plugins

sbt-typesafe-conductr-tester/project/plugins.sbt
```diff
@@ -0,0 +1,2 @@
+lazy val root = Project("plugins", file(".")).dependsOn(plugin)
+lazy val plugin = file("../").getCanonicalFile.toURI
```
huntc: Re-instated the tester - @ktoso it is there for the ease of experimenting during development. Josh showed me this convention a while back.

scripted.sbt
```diff
@@ -1,8 +0,0 @@
-ScriptedPlugin.scriptedSettings
-
-scriptedLaunchOpts := {
-  scriptedLaunchOpts.value ++
-    Seq("-Xmx1024M", "-XX:MaxPermSize=256M", "-Dplugin.version=" + version.value)
-}
-
-scriptedBufferLog := false
```
huntc: Plugin declarations should be within the project folder to keep things tidy.

src/sbt-test/sbt-typesafe-conductr/basic/build.sbt
```diff
@@ -0,0 +1,17 @@
+import ByteConversions._
```
huntc: Renamed the test to "basic" as there will be other tests in the future.

src/sbt-test/sbt-typesafe-conductr/sbt-typesafe-conductr-tester/version.properties
```diff
@@ -1 +0,0 @@
-../../../../version.properties
```
huntc: I really don't get why we went to this property based approach @ktoso - it is more complex than what we have in other plugins such as sbt-bundle.

src/test/scala/com/typesafe/conductr/client/ConductRControllerSpec.scala
```diff
@@ -25,7 +25,7 @@ class ConductRControllerSpec extends WordSpec with Matchers with BeforeAndAfterA
       val testProbe = TestProbe()
       testProbe.send(
         controller,
-        ConductRController.LoadBundle(Uri(BundleFile), None, "some-system", 1.0, 1024, 1024, Set("web-server"))
+        ConductRController.LoadBundle(Uri(testBundle.toURI.toString), None, "some-system", 1.0, 1024, 1024, Set("web-server"))
```
huntc: Test resources have a place - and not at the root of the project.

version.sbt
```diff
@@ -1,12 +1 @@
...
-version in ThisBuild := {
-  val s = new FileInputStream(file("version.properties"))
-  try {
-    val p = new Properties()
-    p.load(s)
-    p.getProperty("version")
-  } finally s.close()
-
-}
+version in ThisBuild := "0.26.0-SNAPSHOT"
```
huntc: @ktoso If you could explain why we went with this property file based approach then that'd be helpful. As you can see what we now have is much simpler.

sbt-typesafe-conductr-tester/project/plugins.sbt
ktoso: I'd usually just cd into one of `src/sbt-test/...` and test there, but no prob if you want the tester around.

.gitignore
ktoso: sorry / thanks for removing

src/sbt-test/sbt-typesafe-conductr/sbt-typesafe-conductr-tester/version.properties
ktoso: I was not aware this is possible in scripted tests:

```
addSbtPlugin("com.typesafe.conductr" % "sbt-typesafe-conductr" % sys.props("project.version"))
```

Would be great if this would have been pointed out during review, would have learnt something about scripted right away :)

version.sbt
ktoso: See https://github.com/sbt/sbt-typesafe-conductr/pull/57#discussion_r27727566 
During review it got a "great", so we merged.

build.sbt
ktoso: didn't know this launchOpts trick, nice.

sbt-typesafe-conductr-tester/project/plugins.sbt
huntc: If you cd to an scripted test then target meta data will be created that can upset scripted. You should not do that.

src/sbt-test/sbt-typesafe-conductr/sbt-typesafe-conductr-tester/version.properties
huntc: Sorry for not feeding back during the review. I want to get your PR in and then tidy-up.

PR #63, opened 2015-04-09
Refactoring the plugin to use single task

### LINKED ISSUE #58: Rationalise to a single conduct command
For consistency with the CLI, and to avoid namespace clash issues with other sbt plugins, we need a single `conduct` command along with `load`, `run` etc as sub commands. One problem with the existing `conductr` configuration approach is that we cannot have a `run` command in place of `start`. `run` is already defined by sbt and has a different meaning. By moving to sub commands then we may have `conduct run` and avoid these style of issues.

Let's also have an `ip` and `port` sub command in place of `controlServer` (we should similarly have an `ip` sub command for the python CLI that sets up the env var accordingly).

eed3si9n:
Fixes #58 

Mostly followed the [plugins best practices](http://www.scala-sbt.org/0.13/docs/Plugins-Best-Practices.html).

- Adds input task `condurctr`, which replaces existing load, start, etc.
- Moves out functionality to plain old Scala object `TypesafeConductR` 
- Adds required plugins `SbtNativePackager && UniversalPlugin && JavaAppPackaging` (Is this enough?)

/review @jsuereth, @huntc


hseeberger:
Nice work!

@huntc, can we use this to rename `TypesafeConductR` to `ConductR`?

huntc:
Thanks @eed3si9n!

Yes, let's drop the "Typesafe" prefix. I now agree that this was overkill. 

Also let's name the "conductr" task just "conduct" as per the Python CLI. 

Heiko there are some style things including using final vals. Small stuff. Could you please review? :-)

hseeberger:
@huntc, I suggest we don't bother Eugene with our specific style guide and instead create an issue and handle renaming and style issues ourselves. WDYT?

huntc:
Sure, whatever works.

PR #65, opened 2015-04-10
Issue # 60. Could not load bundle. Due to the fact that the ConductrCont...

dsugden:
The ConductrController Actor is created onload, with the control server URL as a parameter. This PR adds a message to it's protocol allowing the plugin to change this URL after plugin loading.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/client/ConductRController.scala
+++ b/src/main/scala/com/typesafe/conductr/client/ConductRController.scala
@@ -37,6 +37,12 @@ object ConductRController {
     Props(new ConductRController(address, connectTimeout))
 
   /**
+   * Set the control server host URL
+   * @param host address of the host
+   */
+  case class SetControlServer(host: URI)
+
+  /**
    * Load a bundle with optional configuration.
    * @param bundle The address of the bundle.
    * @param config An optional configuration that will override any configuration found within the bundle.
@@ -163,18 +169,24 @@ class ConductRController(uri: Uri, connectTimeout: Timeout)
   import ConductRController._
   import context.dispatcher
 
-  override def receive: Receive = {
-    case GetBundleInfoStream   => fetchBundleFlow(sender())
-    case request: LoadBundle   => loadBundle(request)
-    case request: StartBundle  => startBundle(request)
-    case request: StopBundle   => stopBundle(request)
-    case request: UnloadBundle => unloadBundle(request)
+  override def receive: Receive = service(uri)
+
+  def service(host: Uri): Receive = {
+    case GetBundleInfoStream   => fetchBundleFlow(sender(), host)
+    case request: LoadBundle   => loadBundle(request, host)
+    case request: StartBundle  => startBundle(request, host)
+    case request: StopBundle   => stopBundle(request, host)
+    case request: UnloadBundle => unloadBundle(request, host)
+    case SetControlServer(uri) =>
+      context.become(service(Uri(uri.toString)))
+      sender() ! "Success"
+
   }
 
   protected def request(request: HttpRequest, connection: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]]): Future[HttpResponse] =
     Source.single(request).via(connection).runWith(Sink.head)
 
-  private def loadBundle(loadBundle: LoadBundle): Unit = {
+  private def loadBundle(loadBundle: LoadBundle, host: Uri): Unit = {
     val bodyParts =
       Source(
         List(
@@ -190,7 +202,7 @@ class ConductRController(uri: Uri, connectTimeout: Timeout)
       )
     val pendingResponse =
       for {
-        connection <- connect(uri.authority.host.address(), uri.authority.port)(context.system, connectTimeout)
+        connection <- connect(host.authority.host.address(), host.authority.port)(context.system, connectTimeout)
         entity <- Marshal(FormData(bodyParts)).to[RequestEntity]
         response <- request(HttpRequest(POST, "/bundles", entity = entity), connection)
         body <- Unmarshal(response.entity).to[String]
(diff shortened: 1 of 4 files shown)


src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -163,18 +169,24 @@ class ConductRController(uri: Uri, connectTimeout: Timeout)
   import ConductRController._
   import context.dispatcher
 
-  override def receive: Receive = {
-    case GetBundleInfoStream   => fetchBundleFlow(sender())
-    case request: LoadBundle   => loadBundle(request)
-    case request: StartBundle  => startBundle(request)
-    case request: StopBundle   => stopBundle(request)
-    case request: UnloadBundle => unloadBundle(request)
+  override def receive: Receive = service(uri)
```
hseeberger: s/uri/host

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -163,18 +169,24 @@ class ConductRController(uri: Uri, connectTimeout: Timeout)
...
+
+  def service(host: Uri): Receive = {
+    case GetBundleInfoStream   => fetchBundleFlow(sender(), host)
+    case request: LoadBundle   => loadBundle(request, host)
+    case request: StartBundle  => startBundle(request, host)
+    case request: StopBundle   => stopBundle(request, host)
+    case request: UnloadBundle => unloadBundle(request, host)
+    case SetControlServer(uri) =>
+      context.become(service(Uri(uri.toString)))
+      sender() ! "Success"
```
hseeberger: "Success" is a bad message. I don't think we need to reply anyway, so please delete and move the `context.become` one line up after `=>`.

src/main/scala/com/typesafe/conductr/sbt/TypesafeConductR.scala
```diff
@@ -30,6 +31,41 @@ private[conductr] object TypesafeConductR {
...
   val actorSystemAttrKey = AttributeKey[ActorSystem]("sbt-typesafe-setConductrTask-actor-system")
 
+  private object IPParser extends RegexParsers {
+
+    /**
+     * http://semberal.github.io/parser-combinators-and-what-if-regular-expressions-are-not-enough.html
+     * @param input
+     * @return
+     */
+    def parseUsingParserCombinator(input: String): Option[String] = {
```
hseeberger: Could you add some tests, please?

hseeberger:
@dsugden, thanks a lot for the contribution! Please take a look at my comments.

dsugden:
@hseeberger Great, Thanks!

I think this PR was premature, I was excited. I did want feedback on this.

I would like to move the IPAddress parsing into the sbt.complete.Parser block in TypesafeConductRPlugin.Parsers as, currently, with this PR, an invalid IP address causes sys error, but this should be caught be parsers.

wrt your first comment above:  

    s/uri/host

I'm not clear on what you intended to tell me.. ? arg naming?

hseeberger:
Great, I already thought it might be better to use sbt parsers instead of Scala ones.

Yes, replace `uri` with `host`.

dsugden:
Issue # 60. 
1) removed scala parser combinators: using sbt.Parser, wrote a naive ipAdress parser that should be sufficient for syntax parsing 
2) changed tab completion for controlServer from http://IP:PORT to IP:PORT
3) still returning a string (hopefully better than "Success") from ConductRController on setting the URL. The ask pattern necessitates this.
4) Added tests for ipAddress parser: changed visibility of Parsers to private[sbt] for the tests

src/main/scala/com/typesafe/conductr/sbt/TypesafeConductR.scala
```diff
@@ -11,16 +11,17 @@ import akka.http.model.{ Uri => HttpUri }
 import akka.pattern.ask
 import akka.util.Timeout
 import com.typesafe.conductr.client.ConductRController
-import com.typesafe.conductr.client.ConductRController.{ LoadBundle, StartBundle, StopBundle, UnloadBundle }
+import com.typesafe.conductr.client.ConductRController._
```
hseeberger: No wildcard imports, please. Maybe you need to reconfigure your IntelliJ?

src/main/scala/com/typesafe/conductr/sbt/TypesafeConductR.scala
```diff
@@ -30,47 +31,58 @@ private[conductr] object TypesafeConductR {
   val conductrAttrKey = AttributeKey[ActorRef]("sbt-typesafe-set-conductr-task")
   val actorSystemAttrKey = AttributeKey[ActorSystem]("sbt-typesafe-setConductrTask-actor-system")
 
+  def setControlServer(host: sbt.URL, s: State, log: Logger): String =
+    withConductRController(s) { conductr =>
+      log.info(s"Setting Control Server URL to $host")
+      val request = SetControlServer(host.toURI)
+      Try(
+        Await.result(conductr.ask(request)(2 seconds).mapTo[String], 2 seconds)
```
hseeberger: Why not use `Await.ready` and return `Unit` instead of the obviously not used string?

src/main/scala/com/typesafe/conductr/sbt/TypesafeConductR.scala
dsugden: yes. done.

src/main/scala/com/typesafe/conductr/sbt/TypesafeConductR.scala
dsugden: no (good) reason. done.

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -163,18 +169,29 @@ class ConductRController(uri: Uri, connectTimeout: Timeout)
...
 
   protected def request(request: HttpRequest, connection: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]]): Future[HttpResponse] =
     Source.single(request).via(connection).runWith(Sink.head)
 
-  private def loadBundle(loadBundle: LoadBundle): Unit = {
+  private def setControlServer(host: URI): Unit = {
+    context.become(service(Uri(host.toString)))
+
+    // TODO verify this is a valid connection?
+    sender() ! "Control Server URL has be set to " + host.toString
```
hseeberger: Sorry for being pesky, but I really think that there's no need to send a reply. Remove these lines and also the use of  the ask pattern in `TypesafeConductR.setControlServer`.

dsugden: all good. I agree. done.

hseeberger:
@dsugden, thanks a lot!!!

huntc:
Thanks Dave!

huntc:
Hmmm. I'm a little concerned with why the previous logic had started to fail...

huntc:
Ah, now I get it - when the `setConductr` input task was introduced it upset the approach we were using, which was to pull the ConductR url from global settings. Given that the previous `conductr` command was a _command_ and not a _task_, the sbt session was updated.

@dsugden I may look at reverting this change, not because it isn't up to scratch or anything; it is great to receive the fix. However I would prefer to get things working the way they were. Instead of a `setConductr` input task I want to revert to the command approach and have `set-conductr-ip` and `set-conductr-port`. These are also analogous to the Python CLI's `CONDUCTR_IP` and `CONDUCTR_PORT` env vars.

dsugden:
All good, this fix was fun for me, my first foray into plugins. Revert away :)

I'm enjoying this, let me know how I can help.

huntc:
Thanks again @dsugden - there's a new PR to replace this: #66

dsugden:
@huntc no worries, just looked at it, I'm intrigued by the sse connection to conductR.
This is an interesting plugin. Looking forward to witnessing this :)

huntc:
The SSE connection is something already handled by the bridge between ConductR and HAProxy (conductr-haproxy). You can subscribe to control protocol events. The Visualizer also uses SSE to update its view of the world. Check out the source in the browser! :-)

PR #78, opened 2015-04-16
Issue # 76

dsugden:
Removed sLog from ConductRPlugin.conductTask. Now using state.log.info(...) for task logging.


huntc:
Thanks!

PR #87, opened 2015-04-28
Downgrade to Play 2.3.8 (closes #86)

### LINKED ISSUE #86: Downgrade to Play 2.3.8

hseeberger:
It looks like in Play 2.4 they have improved the macro for `Json.format` such that it accepts qualified type names like `ConductRController.Attributes`. For Play 2.3 we have to use imports and simple type names like `Attributes`.


hseeberger:
Just released 0.34.0 with these changes.

PR #90, opened 2015-05-01
Pick Configuration at load time. Tab completion for selecting the optional bundle configuration

### LINKED ISSUE #48: Pick configuration at load time, probably also define how ConductR works with "environments"
It's currently rather confusing how to "load bundle with my config", the nearest thing to it docs wise is the section on shazar here: http://typesafehub.github.io/typesafe-conductr/intro/Quickstart.html

One use case is: Apps sometimes have modes, or demo-apps we prepare as activator templates often have `frontend` and `backend` roles (including reactive maps, akkamazing etc). So still while developing the app I want to:
* start bundleX with `backend` configuration on `backend` nodes
* start bundleX with `frontend` configuration on `frontend` nodes.

This is different than ops overlaying their configuration (with for example prod keys / urls etc). I have the full config in my development sources, and it's committed. I'd advocate for supporting "environment configuration", like in chef: https://docs.chef.io/environments.html so we'd have `src/resources/conductr/env.conf` and `src/resources/conductr/end_frontend.conf` etc. 

* I want to be able to pick which env config to shazar and include when loading. It would be nice to have conventions in place, so not every team has it completely differently. Thus the above proposal with `conductr/env[name].conf` (or it could be `.sh`). I'd personally like it to be usable like so: `load -E frontend mybundle`.
* I want to have this name exposed as env variable, so my app can act upon it. We do have activator templates that notice "oh, I'm a frontend node now, okey!".

dsugden:
Fixes #48


src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,6 +45,9 @@ object ConductRPlugin extends AutoPlugin {
       commands ++= Seq(controlServer),
       conduct := conductTask.value.evaluated,
       conductrDiscoveredDist <<= (dist in Bundle).storeAs(conductrDiscoveredDist in Global).triggeredBy(dist in Bundle),
+      conductrDiscoveredConfigDist <<=
+        (dist in BundleConfiguration).storeAs(conductrDiscoveredConfigDist in Global)
+        .triggeredBy(dist in BundleConfiguration),
```
huntc: Style: could you make the `conductrDiscoveredDist` also span multiple lines for consistency?

src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
```diff
@@ -60,11 +63,14 @@ object ConductRPlugin extends AutoPlugin {
...
   private object Parsers {
     lazy val subtask: Def.Initialize[State => Parser[Option[ConductSubtask]]] =
-      Defaults.loadForParser(conductrDiscoveredDist in Global)((state, b) =>
-        (Space ~> (loadSubtask(b) | runSubtask | stopSubtask | unloadSubtask | infoSubtask))?
-      )
-    def loadSubtask(availableBundle: Option[File]): Parser[LoadSubtask] =
-      (token("load") ~> Space ~> bundle(availableBundle) ~ configuration.?) map { case (b, config) => LoadSubtask(b, config) }
+      Defaults.loadForParser(conductrDiscoveredDist in Global) { (state, bundle) =>
+        val conf = conductrDiscoveredConfigDist in Global
+        val maybeBundleConfig = Defaults.loadFromContext(conf, conf.scopedKey, state)
```
huntc: Style:

```scala
        val configSetting = conductrDiscoveredConfigDist in Global
        val bundleConfig = Defaults.loadFromContext(configSetting, configSetting.scopedKey, state)
```

huntc:
Nice!

dsugden:
@huntc I didn't edit the README deliberately. I can take a stab... but figured you'd want to do that.

huntc:
Awesome! Thanks so much @dsugden

PR #92, opened 2015-05-08
Events and logs

2m:
Events and logs JSON format is not yet known, therefore both commands only print a helpful message now.


2m:
Updated with a message print.

PR #94, opened 2015-05-12
The newly introduced spacer caused a transpose error

huntc:
The spacer failed under the circumstances where there were no rows. I suspect it may also have failed for circumstances where there was more than one row also.


src/main/scala/com/typesafe/conductr/sbt/console/Console.scala
```diff
@@ -30,8 +30,7 @@ object Console {
           Name(bundles),
           Replicated(bundles),
           Starting(bundles),
-          Running(bundles),
-          Roles(bundles)
+          Running(bundles)
```
huntc: `Roles` seems to have crept back in...

PR #95, opened 2015-06-02
update to Streams 1.0-RC3

rkuhn:


project/Dependencies.scala
```diff
@@ -2,9 +2,9 @@ import sbt._
 import sbt.Resolver.bintrayRepo
 
 object Version {
-  val akka             = "2.3.10"
-  val akkaContribExtra = "1.16.0"
-  val akkaHttp         = "1.0-RC1"
+  val akka             = "2.3.11"
+  val akkaContribExtra = "1.17.0-SNAPSHOT"
```
rkuhn: this will need to be corrected once akkaContribExtra is published

PR #103, opened 2015-09-15
New sbt-bundle and protocol

huntc:
Caters for a new protocol format and the delivery of addition multipart scheduling fields. Defaults though to protocol 1.0 which is required by ConductR 1.0 and supported by ConductR 1.1.


src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -179,15 +182,18 @@ class ConductRController(conductr: Uri, loggingQuery: Uri, connectTimeout: Timeo
...
+    case request: GetEventStream      => fetchLoggingQuerySource[Event](sender(), request.protocolVersion, request.bundleId)
+    case request: GetLogStream        => fetchLoggingQuerySource[Log](sender(), request.protocolVersion, request.bundleId)
+    case request: LoadBundle          => loadBundle(request)
+    case request: RunBundle           => runBundle(request)
+    case request: StopBundle          => stopBundle(request)
+    case request: UnloadBundle        => unloadBundle(request)
   }
 
+  private def protocolVersionPath(version: String): String =
+    if (version == "1.0") "" else "/v" + version
```
markusjura: Can't we have the same version scheme for `1.0` and future versions:

```
v1.0
v1.1
v1.2
```

huntc: That would mean creating extra routes in ConductR - also if we issue a `/v1.0` to ConductR 1.0 then it will fail.

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -207,24 +213,37 @@ class ConductRController(conductr: Uri, loggingQuery: Uri, connectTimeout: Timeo
...
-        FormData.BodyPart.Strict("memory", loadBundle.memory.toString),
-        FormData.BodyPart.Strict("diskSpace", loadBundle.diskSpace.toString),
-        FormData.BodyPart.Strict("roles", loadBundle.roles.mkString(" ")),
-        FormData.BodyPart.Strict("bundleName", toBundleName(loadBundle.bundle)),
-        bundleFileBodyPart
-      ) ++ configFileBodyPart
-    )
+    configFileBodyPart: Option[FormData.BodyPart]): Source[FormData.BodyPart, Unit] = {
+    val schedulingRequirements = loadBundle.protocolVersion match {
+      case "1.0" =>
```
markusjura: The valid version numbers should be factored out to constants or ADTs.

markusjura:
What about renaming `protocolVersion` to `apiVersion`? This is the typically naming for versions in an REST API.

src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
```diff
@@ -25,25 +26,36 @@ private[conductr] object ConductR {
   final val DefaultConductrProtocol = "http"
   final val DefaultConductrHost = "127.0.0.1"
   final val DefaultConductrPort = 9005
+
   val conductrAttrKey = AttributeKey[ActorRef]("sbt-conductr")
   val actorSystemAttrKey = AttributeKey[ActorSystem]("sbt-conductr-actor-system")
 
-  def loadBundle(bundle: URI, config: Option[URI], loadTimeout: Timeout, state: State): String = {
+  def loadBundle(pv: String, bundle: URI, config: Option[URI], loadTimeout: Timeout, state: State): String = {
```
markusjura: `protocolVersion` instead of `pv` (Naming should be consistent across the project).

src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
```diff
@@ -25,25 +26,36 @@ private[conductr] object ConductR {
...
     def get[A](key: SettingKey[A]) =
-      Project.extract(state).getOpt(key)
+      getOpt(key)
         .fold(Bad(One(s"Setting ${key.key.label} must be defined!")): A Or One[String])(Good(_))
-    def doLoadBundle(stm: String, roles: Set[String], nrOfCpus: Double, memory: Bytes, diskSpace: Bytes) = {
+    def getOpt[A](key: SettingKey[A]) =
+      Project.extract(state).getOpt(key)
+    def doLoadBundle(nn: String, v: String, stm: String, roles: Set[String], nrOfCpus: Double, memory: Bytes, diskSpace: Bytes) = {
```
markusjura: What is `nn` and `v`? Better naming?

huntc:
Changed to api version

src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
huntc: kind of in the spirit of what was already there...

markusjura: Well, `roles` or `nrOfCpus` are self explaining. But not `v` or `nn`. At least not for me :-)

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
markusjura: Ah, I was misreading the logic. Thought that we return `1.0` if `CondcutRPlugin.conductrApiVersion` is set to `1.0`. Instead we return an empty string in this case which makes sense.

markusjura:
LGTM (Beside the comments).

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
huntc: Wilco

src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
huntc: The problem with that here is that you'll hide the sbt definition or `normalizedName`. `nn` == `normalizedName`.

huntc: ..as in you'll conflict with the sbt settings themselves.

markusjura: The naming should be consistent across the sbt definition and method parameters across all files in ConductR. So either we use `av` or `apiVersion` all over the place.

huntc: I've got another idea...

src/main/scala/com/typesafe/conductr/sbt/ConductRKeys.scala
```diff
@@ -16,5 +16,6 @@ trait ConductRKeys {
   val conductrConnectTimeout = settingKey[Timeout]("The timeout for ConductR communications when connecting")
   val conductrLoadTimeout = settingKey[Timeout]("The timeout for ConductR communications when loading")
   val conductrRequestTimeout = settingKey[Timeout]("The timeout for ConductR communications when requesting")
+  val conductrApiVersion = settingKey[String]("The protocol to use when communicating with ConductR. Defaults to 1.0 which is required by ConductR 1.0.")
```
markusjura: The api version to use when communcating..

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
```diff
@@ -26,11 +26,15 @@ import play.api.libs.json.Reads
 
 object ConductRController {
 
-  private[this] val BundlePath = "(.*?)(-[a-fA-F0-9]{0,64})?\\.zip".r
+  /*
+   * Supported ConductR Control Protocol versions.
+   */
+  final val ApiVersionV10 = "1.0"
```
markusjura: What about putting that into a separate scala class / file under the `client` package:

```
object ApiVersion {
  trait Version
  case class V10 extends Version
  case class V11 extends Version
}
```

markusjura: Either as an ADT or pure constants.

src/main/scala/com/typesafe/conductr/sbt/ConductRKeys.scala
huntc: well... ok

markusjura: Also need to be changed in the README

src/main/scala/com/typesafe/conductr/client/ConductRController.scala
huntc: OK

fsat:
@hutnc - LGTM - I should be able to translate this to CLI.

PR #104, opened 2015-09-20
Implementation of events and logs commands

markusjura:
Implementation of the commands `events` and `logs`. Syntax:

```
conduct logs visualizer
conduct logs conductr-kibana --scale 50
conduct events visualizer
conduct events conductr-kibana --scale 50
```

The last parameter 50 is optional and tells how many lines should be displayed.

The logs and events are getting retrieved. However, the PR isn't finished yet. Missing pieces are:
- Nice formatting of output

Also refactored the `src/main/scala/com/typesafe/conductr/client/ConductRController.scala` by encapsulating duplicated code into methods.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
@@ -6,11 +6,16 @@ package com.typesafe.conductr.sbt
 package console
 
 import com.typesafe.conductr.client.ConductRController
+import jline.TerminalFactory
+import org.joda.time.DateTime
 
 object Column {
 
   import AnsiConsole.Implicits._
   import Console.Implicits._
+
+  private final val Space = " "
+  private final val NewLine = "\n"
 
   /**
    * A fixed width column with no additional functionality.
@@ -60,24 +65,46 @@ object Column {
     protected def justify(s: String): String
 
     protected def space(length: Int): String =
-      Stream.continually(" ").take(length).mkString
+      Stream.continually(Space).take(length).mkString
 
     /**
      * Add '... ' if string is over the allowed width.
      */
-    private def ellipsize(s: String): String = {
+    private[console] def ellipsize(s: String): String = {
       val ellipsis = "... "
-      if (s.visibleLength >= width)
+      if (s.visibleLength > width)
         s.take(width - ellipsis.length) + ellipsis
       else
         s
     }
   }
 
-  trait LeftJustified { self: RegularColumn =>
+  /**
+   * The wrapping column wraps the text instead of cutting it based on the column length.
+   * If the text is longer than the width of the column, one additional line as separation are added. This
+   * new line is necessary so that the next row is displayed on a new line.
+   */
+  trait WrappingColumn extends RegularColumn {
+
+    /**
+     * Left margin. If the text character count is greater than the column with this margin size
+     * is used to align the the second line of the text within the column.
+     */
+    def marginLeft: Int
+
     override protected def justify(s: String): String = {
-      s + space(width - s.visibleLength)
+      val formattedText =
+        if (s.visibleLength > width) s.grouped(width).mkString(Space * marginLeft) else s
+      super.justify(formattedText)
     }
+
+    override private[console] def ellipsize(s: String): String =
+      if (s.visibleLength > width) s + NewLine else s
+  }
+
+  trait LeftJustified { self: RegularColumn =>
+    override protected def justify(s: String): String =
+      s + space(width - s.visibleLength)
   }
 
   trait RightJustified { self: RegularColumn =>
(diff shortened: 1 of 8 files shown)


huntc:
Looking good.

markusjura:
Alrighty, the format has been optimized for the `logs` and `events` command. In case the log or event message length is less than the column width it will look like this:

screen shot 2015-09-22 at 15 23 05

If the length of the text is longer than the column width we will wrap the text and add at the end of the message an additional new line for readability reasons:

screen shot 2015-09-22 at 15 24 12

**Note:** The last character of the log messages of `conduct logs` is currently a new line. This will be fixed in another PR.

markusjura:
@huntc: The travis test is failing because it runs the tests with Java 7, but I've used the `java.time.*` objects introduced in Java 8.

Should we support Java 7 for `sbt-conductr` as well? I guess if conductr has Java 8 as a prerequisite itself we should we fine to use Java 8 objects in `sbt-conductr`. WDYT?

huntc:
The plugins themselves require Java 7 to support Play 2.3 and Akka 2.3 users. It would have been Java 6 but the native packager brought us to 7.

src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
```diff
@@ -65,19 +69,33 @@ object Column {
     /**
      * Add '... ' if string is over the allowed width.
      */
-    private def ellipsize(s: String): String = {
+    protected def ellipsize(s: String): String = {
```
huntc: Why the scope change? If it is required by a test then `private[console]`.

src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
```diff
@@ -65,19 +69,33 @@ object Column {
...
+   * new line is necessary so that the next row is displayed on a new line. The second new line has been added
+   * for readability reasons. On the terminal screen only one line as a separation is visible.
+   */
+  trait WrappingColumn extends RegularColumn {
+    override protected def ellipsize(s: String): String =
+      if (s.visibleLength > width)
+        s + "\n\n"
+      else
+        s
+  }
```
huntc: Could we also align the wrapped text with the column? :-)

markusjura: Not sure what you mean with that.

src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
markusjura: Needed to be changed because we override this method in sub traits. Changed it to `private[console]`.

markusjura:
Used `joda-time` 2.8.2 instead of `java.time.*` to be Java 7 compatible.

src/main/scala/com/typesafe/conductr/sbt/console/Column.scala
huntc: Simply that when we wrap, instead of starting the next line in the first column position, we start it aligned with the start of the column.

i.e. not

```
col1 col2
     this is some
text that flows
```

...but

```
col1 col2
     this is some
     text that flows
```

markusjura: Got it. Then we need to change it in `conductr-cli` as well. We do not align it there as well.

huntc: My view is that the python cli should not worry about truncating nor wrapping - just output. Use tools like `more` to deal with that by piping the output, using `sed` etc.

markusjura: Ok, makes sense.

typesafe-tools:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;8meDbFgO&#x2F;113-sbt-conductr-needs-to-enable-the-logs-and-events-command"> sbt-conductr needs to enable the logs and events command</a>

markusjura:
Log messages will look now like this:

screen shot 2015-09-28 at 11 55 44

- The text wrapping takes place only in the column itself and does not expand to other columns.
- No new lines in between are therefor necessary

huntc:
Looking good! 

How come the msg isn't being parsed for those columns being wrapped though?

markusjura:
With parsing you mean that we should, e.g. apply json parsing if the log message has a json format? So far no parsing is applied. Only formatting, so that the wrapping only happens within the same column.

huntc:
Sorry for any confusion. I'm curious as to why the multiline logs are showing as JSON, but the others don't. Is this the same output as the python CLI?

src/main/scala/com/typesafe/conductr/sbt/console/Screen.scala
```diff
@@ -52,7 +52,7 @@ class Screen[A <: Iterable[B], B](refresh: Boolean, layout: A => Screen.Layout)
     case CheckSize =>
       checkSize()
 
-    case Status.Failure(_) =>
+    case Status.Failure(ex) =>
```
huntc: ex is unused

markusjura:
Kibana always produces JSON as log messages. The first 3 lines are part of the Kibana start script.

It is the same output as on the python CLI.

markusjura:
Removed the `ex` from the `Status.Failure`. Ready to get merged?

huntc:
Thanks for the explanations. Lgtm.

Issue #102, opened 2015-09-11
Wrong conduct info output


markusjura:
The command `conduct info` interferes with the next terminal line:

```bash
> conduct info
[success] Total time: 0 s, completed Sep 11, 2015 9:22:21 AMREP   #STR   #RUN
> 5172d     visualizer                 1      0      1
1fecf0e     conductr-elasticsearch     1      0      1
6a043df     conductr-kibana            1      0      1
```

Instead, first the entire content of `conduct info` should be displayed as output and afterwards the pointer for the next terminal commands should be displayed.

PR #108, opened 2015-09-28
Fix wrong cursor position in conduct info, events, logs

markusjura:
Fixes issue #102.

So far the cursor position has been saved and restored after the output has been rendered. This resulted into having the sbt console cursor inside of the output. 

Not sure why this logic has been applied in the first place.
@huntc: Do you have any insights here?

The refresh mode is currently not used by any of the commands.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/console/Console.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/console/Console.scala
@@ -92,14 +92,14 @@ object Console {
 
     import system.dispatcher
 
-    print(Ansi.ansi().saveCursorPosition())
-
     if (refresh) {
+      print(Ansi.ansi().saveCursorPosition())
       val con = new ConsoleReader()
       blocking {
         con.readCharacter('q')
       }
       system.stop(screen)
+      print(Ansi.ansi().restorCursorPosition())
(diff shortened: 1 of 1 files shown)


huntc:
What is the effect of this when running conduct info in continuous mode?

huntc:
Hmm. Refresh used to be supported by "conduct info". That's why the saving of cursor position occurred. I'll take a look.

markusjura:
By default refresh is set to false: https://github.com/sbt/sbt-conductr/blob/master/src/main/scala/com/typesafe/conductr/sbt/ConductR.scala#L143

It also didn't get overriden: https://github.com/sbt/sbt-conductr/blob/master/src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala#L141

huntc:
OK, so can we keep it in there for when refresh is enabled for the future...?

markusjura:
Yes, should I move the `saveCursorPosition` and `restorCursorPosition` into the refresh logic?

huntc:
That would be great - thanks

markusjura:
Done. Mergeable?

huntc:
Nice one. LGTM.

PR #109, opened 2015-10-01
Merge bundle configurations

markusjura:
This PR changes the `load bundle` logic. Instead of constructing the `LoadBundle` infos from the`BundleKeys`, the information are retrieved from the `bundle.conf`inside the bundle zip file.

If an additional configuration zip file (`configuration:dist`) is specified during `bundle load` then the infos of the `bundle.conf` inside the configuration zip file are used. These infos are overriding the values of the bundle.

This is the same logic as in the `conductr-cli`. It has been manually tested with and without additional configuration. Unit tests have not been provided, yet. First, @huntc should review the current approach. Afterwards I'll create tests for it.

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
@@ -4,7 +4,9 @@
 
 package com.typesafe.conductr.sbt
 
+import java.io.{ IOException, InputStreamReader }
 import java.net.URL
+import java.util.zip.{ ZipException, ZipFile }
 
 import akka.actor.{ ActorRef, ActorSystem }
 import akka.http.scaladsl.model.{ Uri => HttpUri }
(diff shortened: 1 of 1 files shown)


src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
```diff
@@ -30,18 +34,39 @@ private[conductr] object ConductR {
...
+        mergeConfigKey[String](baseConf, overlayConfOpt, _.getString("system")),
+        mergeConfigKey[Set[String]](baseConf, overlayConfOpt, _.getStringList("roles").asScala.toSet),
+        mergeConfigKey[Double](baseConf, overlayConfOpt, _.getDouble("nrOfCpus")),
+        mergeConfigKey[Long](baseConf, overlayConfOpt, _.getLong("memory")),
+        mergeConfigKey[Long](baseConf, overlayConfOpt, _.getLong("diskSpace")))
+    def mergeConfigKey[A](baseConf: Config, overlayConfOpt: Option[Config], f: Config => A): A =
+      overlayConfOpt match {
+        case None              => f(baseConf)
+        case Some(overlayConf) => Try(f(overlayConf)).getOrElse(f(baseConf))
+      }
```
huntc: Why have all this merge logic when typesafe config can merge for us? HOCON can be merged in many different ways - more sophisticated that what we have here.

Other than that, the approach looks reasonable.

markusjura: Oh, was not aware of this. I'll change it accordingly.

markusjura:
The two `bundle.conf`s are now merged with the Typesafe Config method `withFallback`.

huntc:
This looks really, really nice.

I realise that we don't have a great number of scripted tests for it, but some tests exercising this would be awesome.

Otherwise LGTM.

markusjura:
I'll add these tests as "acceptance tests" similar to the logs and events tests. Reason is we can't test the new behaviour based on how we currently do tests, by mocking the respective actor. Now we need a bundle and configuration zip file to test the unzip and merging logic. I think for this case an acceptance test is better here. Will add them as soon as we have the testkit in place.

markusjura:
Added the issue #110 accordingly to document the necessary tests.

Issue #97, opened 2015-08-27
Not handling ambiguous replies properly


huntc:
It appears that the http "ambiguous" status code is not handled properly:

```
[error] (*:conduct) com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'Specified': was expecting ('true', 'false' or 'null')
[error]  at [Source: java.io.StringReader@2ec32cae; line: 1, column: 10]
```

huntc:
@markusjura - can you please have a look at this given that you're going to be in there?

markusjura:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;2aYWyvJb&#x2F;121-tactical-activities"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Tactical Activities</a>

markusjura:
Fixed with PR #111.

PR #111, opened 2015-10-04
Throw error if http status code 300

markusjura:
ConductR is returning the status code 300 '300 Multiple Choices' in case the http request is ambiguous. in Akka HTTP the status code 300 is part of the `redirection` group. The redirection group returns `true` for the method `status.isSuccess`. In `sbt-conductr` we only validated the status code based on the method `status.isSuccess`. Now an additional check has been added which is validating if the status code is 300. If that is the case an `IllegalStateException` is thrown, as with other errors.


markusjura:
Fixes issues #97.

PR #117, opened 2015-10-23
Resolve conductr host address more intelligent

markusjura:
Instead of using `127.0.0.1` as the default conductr host ip address the following logic is now executed:
1. Get docker host ip address with `docker-machine` (sbt-condutr-sandbox scenario). If `docker-machine` does not return an ip address:
  - Get docker host ip address with `boot2docker` (sbt-condutr-sandbox scenario). If `boot2docker` does not return an ip address:
    - Get ip address by running `hostname`.

Fixes issue https://github.com/sbt/sbt-conductr/issues/116.


markusjura:
The calculation of the ip address is encapsulated in the method `ConductRUtils. resolveDefaultHostIp`. This method is available for every sbt plugin which depends on `sbt-conductr`. Therefor this method can also be used in the `conductr` project in the acceptance tests.

huntc:
Can we please think for a better name than utils? These become dumping grounds otherwise.

markusjura:
It is a "dumping ground for all publicly available methods". What about using `ConductR` and replacing the current ConductR object with `ConductRService`.

As you can see I am really bad in naming things ;-)

huntc:
Why not make it a public function on the plugin itself?

markusjura:
Alright, let's do that.

markusjura:
Done

PR #118, opened 2015-10-23
Adding usage info to conduct commands

markusjura:
The conduct commands provide now helpful usage information in case the command was not used correctly:
- `conduct` | `conduct help` => Usage information of conduct command
- `conduct [subtask] => Usage information of the particular subtask (only if the parsing of the command failed)

The usage information are the same as on `conductr-cli`.


huntc:
Nice!

PR #121, opened 2015-10-29
Better error handling to resolve host ip

markusjura:
The current error handling of `docker-machine` is not optimal. If the VM has not been started with `docker-machine start default` and the user start the sbt session then he will only see this output currently:

```
...
host is not running
[info] Set current project to sbt-conductr-tester (in build file:/Users/mj/workspace/sbt-conductr/sbt-conductr-tester/)
[info] Control Protocol set for http://:9005. Use 'controlServer {ip-address}' to set an alternate address.
```

The error message `host is not running` comes straight from docker machine. The user doesn't now what to do here. This PR improves the output if the VM has not been started yet with `docker-machine`. The new output is:

```
...
host is not running
[warn] Docker VM has not been started. Use 'docker-machine start default' in the terminal to start the VM and reload the sbt session.
[info] Set current project to sbt-conductr-tester (in build file:/Users/mj/workspace/sbt-conductr/sbt-conductr-tester/)
[info] Control Protocol set for http://:9005. Use 'controlServer {ip-address}' to set an alternate address.
```

The new warning is color coded with `println` instead of using `state.log` logger. The logger can't be used here because the method `resolveDefaultHostIp` is called when no state is available (e.g. in ConductR acceptance tests).

The error handling for `boot2docker` stays the same. If both `docker-machine` and `boot2docker` and the `hostname` command doesn't exist then the default ip address `127.0.0.1` is returned. 

@huntc: Instead of returning `127.0.0.1` we could also throw a warning saying that docker need to be installed. This case can only occur in windows scenarios.

@huntc: Can you do a review?

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/ConductR.scala
@@ -27,7 +27,7 @@ private[conductr] object ConductR {
   import Import._
 
   final val DefaultConductrProtocol = "http"
-  final val DefaultConductrHost = ConductRPlugin.resolveDefaultHostIp
+  final val DefaultConductrHost = ConductRPlugin.resolveDefaultHostIp()
   final val DefaultConductrPort = 9005
 
   val conductrAttrKey = AttributeKey[ActorRef]("sbt-conductr")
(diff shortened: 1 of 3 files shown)


huntc:
Lgtm. Agree re 127.0.0.1

Issue #114, opened 2015-10-20
Improve error messages for logs, events, services, info


markusjura:
The error message for the following conduct commands need to be improved:
- logs
- events
- services
- info

So far these commands are timing out after 5 seconds if ConductR does not respond, e.g. if ConductR has not been started. Especially as a first user experience this is particularly bad. Instead the failure case should return with a concrete error message.

This issue should be fixed while implementing the new HTTP library for `sbt-conductr`.

markusjura:
To be more concrete, the commands should:
- Try to fail fast. If control server is not set the current command should set the control server
  - If setting the control server fails it should display an error message that docker has not been started (ip address is not reachable)
- In all cases it should return the most concrete error message

@huntc: WDYT?

huntc:
Sounds good

markusjura:
Addressed by PR https://github.com/sbt/sbt-conductr/pull/124.

PR #123, opened 2015-11-10
WIP - DON'T MERGE: Split out sbt-conductr-client

markusjura:
Splitting out the `client` package into a separate library. This is necessary in order to use the control protocol standalone in other projects, e.g. for a ConductR testkit.

This PR also aims to improve the client library in order to fix:
https://github.com/sbt/sbt-conductr/issues/114
https://github.com/sbt/sbt-conductr/issues/83


markusjura:
@huntc: Can you check if this goes into the right direction. I've splitted up `sbt-conductr` into 2 sub modules:
- client: contains the `ConductRController`, i.e. it sends HTTP requests to the ConductR control server.
- sbt-conductr: Handles the sbt related stuff

Tomorrow I'll fix the issues #114 and #82. 

This PR needs to be merged after @fsat has merged the PR https://github.com/sbt/sbt-conductr/pull/122.

sbt-conductr/src/main/resources/LICENSE-2.0.txt
```diff
@@ -0,0 +1,202 @@
+
```
huntc: Why is this license required here i.e. as part of resources? We have a LICENSE file which is all that is required.

huntc:
The client library feels like something that can potentially have its own release cycle and thus independent of sbt-conductr. For example I can see that it will be used by our test kit. The test kit should have no direct dependency on sbt and could be used outside of an sbt context.

I recommend a separate repository named `conductr-client`.

markusjura:
Duplicate to https://github.com/sbt/sbt-conductr/pull/124

PR #124, opened 2016-01-13
Integrate conductr-client-lib

markusjura:
Uses the new [conductr-clientl-lib](https://github.com/typesafehub/conductr-bundle-lib) to interact with the ConductR REST API. The `ConductRController` has been deleted. This PR removed the support for ConductR's API version 1. Only version 2 is supported.

Also this PR fixes several issues:
- https://github.com/sbt/sbt-conductr/issues/114: `conduct events`, `conduct logs` and `conduct info` now fail fast in case the connection couldn't be established instead of waiting 5 seconds. Also the error message of the HTTP response are now returned to the sbt console to provide more useful information
- https://github.com/typesafehub/conductr/issues/590: If the bundle is too large then loading a bundle will return an HTTP error which is now forwarded to the sbt console
- The timestamp of `conduct events` and `conduct logs` was displayed as `hh:mm:ss` instead of `HH:mm:ss`. `hh` uses the 12h format whereas `HH` uses the 24h format. So 6:30pm was displayed as "06:30:00" and is now displayed as "18:30:00".

This PR can't be merged yet because the conductr-client-lib version 1.2.0 is not released yet.


markusjura:
Currently changing the tests accordingly.. With that the code changes a bit as well. Code review should not be done yet.

markusjura:
Alright, new tests have been pushed. I've tested this PR with my local conductr-client-lib version and everything worked fine. All tests are passing as well (with the local conductr-client-lib SNAPSHOT).

@huntc: Can you please do a review.

huntc:
LGTM

PR #127, opened 2016-03-10
Use conductr-cli commands

markusjura:
The [conductr-cli](https://github.com/typesafehub/conductr-cli) and sbt-conductr using the exact same commands to communicate with the ConductR cluster. However, both libraries do not share any code so far. This PR removes the sbt-conductr implementation of the `conduct` commands and instead calls the `conductr-cli`.

**Pros**
- Same codebase results into less work when adding new features or changing existing commands
- Same codebase means that bots libs have the same amount of commands and the same command syntax
- `conduct load`, `conduct run`, etc. are waiting now until the bundle has been loaded / started (current logic from conductr-cli)
- `conduct load` can now resolve bundles from bintray (current logic from conductr-cli)

**Cons**
- The `conductr-cli` need to be installed to use `sbt-conductr`

**Changes**
- ConductR client implementation has been removed
- The following `ConductRKeys` have been removed. Instead the CLI options should be used. Note that if no option is set then the default from the `conductr-cli` is used:
  - `controlServer`
  - `conductrConnectTimeout`
  - `conductrLoadTimeout`
  -.`conductrRequestTimeout`
  - `conductrControlServerUrl`
  - `conductrLoggingQueryUrl`
  - `conductrApiVersion`
- `conduct services` command has been added
- The conductr-cli help for each sub command is now available in sbt-conductr as well, e.g. `conduct load --help`
- All common options from the `conductr-cli` have been added to `sbt-conductr` as well (e.g. `--verbose`, `--long-ids`)
- The order of arguments have been changed from `conduct <command> <positional args> <optional args>` to `conduct <command> <optional args> <positional args>`. Example `conduct run --verbose --scale 2 my-bundle`. Note that the auto completion for a bundle (conduct run, stop, etc.) and for bundle files (conduct load) still works. As a consequence we don't have auto completion for any optional argument anymore (e.g. `--scale`, `--verbose`, etc.)
- Added a validation check if the `conductr-cli` is installed. If not then an `sys.error` is thrown with instructions how to install the CLI. This check is executed for every `conduct` command


markusjura:
Added the issue https://github.com/sbt/sbt-conductr/issues/128 to update the README and provide `sbt scripted` tests.

src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
```diff
@@ -26,33 +21,21 @@ object ConductRPlugin extends AutoPlugin {
...
-      ConductRKeys.conductrLoggingQueryUrl := envUrl("LOGGING_QUERY_IP", DefaultConductrHost, "LOGGING_QUERY_PORT", DefaultConductrPort, DefaultConductrProtocol),
-
-      ConductRKeys.conductrApiVersion := "1"
+      dist in BundleConfiguration := file("")
     )
 
   override def projectSettings: Seq[Setting[_]] =
     super.projectSettings ++ List(
-      Keys.commands ++= Seq(controlServer),
+      Keys.commands ++= Seq.empty,
```
huntc: Why not just remove this line?

PR #130, opened 2016-03-11
Auto complete bundle names

markusjura:
For the following commands the available bundle names in the ConductR cluster are resolved and auto completed if the user presses the tab key:
- conduct run
- conduct stop
- conduct load
- conduct unload
- conduct logs
- conduct events

Fixes https://github.com/sbt/sbt-conductr/issues/43

The current solution runs `conduct info` to resolve the current bundle names of the running ConductR cluster. If no ConductR cluster is running then the exception if catched and an empty String is returned, meaning no auto completion is available. What is not so nice is that the `conduct info` command is executed every time the user hits the tab key when using the `conduct` command, i.e. `conduct info` is executed for `conduct run` as well as for `conduct load`, `conduct --help` and so on. This slows down the auto completion at about 0.5 seconds for the commands which are not interested in the bundle names.

@huntc: I couldn't find a better scope to execute [conduct info](https://github.com/markusjura/sbt-conductr/blob/b8bdc81979ce30f7bf923b74dbb8a4c016ac8a7e/src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala#L69). Do you have an idea how to improve this?

### DIFF
--- a/src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
+++ b/src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
@@ -47,26 +47,30 @@ object ConductRPlugin extends AutoPlugin {
 
   private object Parsers {
     lazy val subtask: Def.Initialize[State => Parser[ConductSubtask]] = {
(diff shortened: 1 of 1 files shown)


huntc:
LGTM

PR #135, opened 2016-03-30
Sandbox CLI support

markusjura:
Adding support for the CLI `sandbox` command.

This PR makes [sbt-conductr-sandbox](https://github.com/typesafehub/sbt-conductr-sandbox) obsolete. Instead of re-implemeting the sandbox in sbt, we now make use of the sandbox command of the CLI. This has the following consequence compared to using sbt-conductr-sandbox:
- Instead of defining sbt settings keys the user now need to specify arguments when starting the sandbox (as in the CLI). The following sbt settings keys have been removed:
  - SandboxKeys.imageVersion
  - SandboxKeys.image
  - SandboxKeys.envs
  - SandboxKeys.ports
  - SandboxKeys.debugPort
  - SandboxKeys..logLevel
  - SandboxKeys..conductrRoles
- The `sandbox debug` command and with it all debug functionality has been removed
- It is not possible anymore to specify the `SandboxKeys.imageVersion` in the `build.sbt`. Instead the user need to specify the version in the `sandbox run` command or the project needs to be an RP project. In the latter case the latest ConductR image version is used (same logic as before)

**Notes**
- sbt scripted tests have been added to validate:
  - All sandbox arguments
  - Bundle ports + sandbox run arg ports
  - Scaling
 - Features
 - RP project logic
- Parsing logic has been generally improved and the `sandbox` and `conduct` parsing is now very much the same


markusjura:
@huntc @fsat: Please review :-)

Is there any change necessary due to the ACL changes (e.g. in the sbt scripted tests)?

src/main/scala/com/typesafe/conductr/sbt/ConductRPlugin.scala
```diff
@@ -20,217 +22,67 @@ object ConductRPlugin extends AutoPlugin {
   import sbinary.DefaultProtocol.FileFormat
 
   val autoImport = Import
+  import ConductRKeys._
```
huntc: In general, I've always found it better to be specific about which key I want to use within a plugin in order to avoid namespace clashes. Just a FYI, not terribly important.

huntc:
Nice one!

PR #140, opened 2016-04-01
Ensure correct acls declaration within bundle.conf is generated correctly when RequestMapping is empty

fsat:


src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -306,10 +306,18 @@ object BundlePlugin extends AutoPlugin {
...
+      acl.protocolFamilyRequestMappings
+        .map(formatRequestMappings)
+        .mkString(
+          "          {",
+          ",\n",
+          "          }"
+        )
+    else
+      ""
+  }
```
huntc: No braces. Also use fold?

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -307,9 +307,12 @@ object BundlePlugin extends AutoPlugin {
...
   private def formatAcl(acl: RequestAcl): String =
-    s"""          {
-        |${acl.protocolFamilyRequestMappings.map(formatRequestMappings).mkString(",\n")}
-        |          }""".stripMargin
+    if (acl.protocolFamilyRequestMappings.isEmpty)
+      ""
+    else
+      s"""          {
+          |${acl.protocolFamilyRequestMappings.map(formatRequestMappings).mkString(",\n")}
+          |          }""".stripMargin
```
fsat: @huntc - brace removed, can't use `fold` since `acl.protocolFamilyRequestMappings` is a `Set[ProtocolFamilyRequestMapping]`

PR #141, opened 2016-04-03
Integrate sbt-lagom-bundle

markusjura:
Adds the auto plugin from `sbt-lagom-bundle` into `sbt-conductr`. This means whenever the `sbt-conductr` library is added then the `LagomBundlePlugin` is triggered for every `LagomPlay` project.

**Note**
- Now `sbt-conductr` is adding the `lagom-sbt-plugin` library as a dependency. This library is now part of the classpath, also for projects that are adding `sbt-conductr` but are not a Lagom project.


huntc:
Lgtm

Not so worried about Lagom being added - no more so than Play being added. The classpath being added to is the build classpath, and not the project's. This means extra dependencies if you're building a pure Scala/Java/Akka app, but hopefully this isn't considered a burden. The benefits of a single pluggin being added should trump that hopefully.

markusjura:
Well said. Agree!

PR #146, opened 2016-04-05
Lagom play bundle

markusjura:
Introduces a new bundle plugin that get triggered for projects that have `LagomPlay` enabled, i.e. a bundle plugin for Play projects in Lagom.

This new plugin sets default bundle keys and adds the lagom conductr-bundle-lib to the project.


markusjura:
This PR also makes sure that the particular bundle plugins (e.g. `LagomBundlePlugin`) is only triggered if the corresponding plugin (e.g. `LagomJava`) is not only part of the classpath, but also enabled by the user.

src/main/scala/com/lightbend/conductr/sbt/LagomBundlePlugin.scala
```diff
@@ -11,6 +11,42 @@ import play.api.libs.json._
...
+  import PlayBundleImport._
+  import BundlePlugin.autoImport._
+
+  val autoImport = LagomBundleImport
+
+  private val classLoader = this.getClass.getClassLoader
+
+  override def requires =
+    withContextClassloader(classLoader) { loader =>
+      Reflection.getSingletonObject[Plugins.Basic](classLoader, "com.lightbend.lagom.sbt.LagomPlay$") match {
```
huntc: I like it

src/main/scala/com/lightbend/conductr/sbt/PlayBundlePlugin.scala
```diff
@@ -16,8 +16,12 @@ object PlayBundlePlugin extends AutoPlugin {
...
   override def requires =
     withContextClassloader(classLoader) { loader =>
-      val isPlayProject = (Reflection.getSingletonObject[Plugins.Basic](classLoader, "play.sbt.Play$") orElse Reflection.getSingletonObject[Plugins.Basic](classLoader, "play.Play$")).isSuccess
-      if (isPlayProject) BundlePlugin else NoOpPlugin
+      def play23OrBelow = Reflection.getSingletonObject[Plugins.Basic](classLoader, "play.Play$")
+      def play24OrAbove = Reflection.getSingletonObject[Plugins.Basic](classLoader, "play.sbt.Play$")
+      play24OrAbove orElse play23OrBelow match {
+        case Failure(_)    => NoOpPlugin
+        case Success(play) => BundlePlugin && play
+      }
```
huntc: Very nice

PR #147, opened 2016-04-10
Support acl and service endpoints

markusjura:
This PR changes a couple of things in respect to support ACL and service endpoints
- Changing `BundleConfVersion` notation to `V1`, `V2`. Currently only `V1` is allowed.
- Introduction of the setting `condutrTargetVersion` to validate the settings against a certain ConductR version
- Introduction of the setting flag `enableAcls`. This flag is used to indicate if acls can be used at the endpoint declaration. If set to true then the Lagom endpoints will be declared as acl endpoints, otherwise as service endpoints
- Introduction of a setting validation. This validation is triggered whenever a `bundle:stage`, `bundle:dist`, `configuration:stage` or `configuration:dist` command is executed. All validation checks are processed and in case one or multiple checks have failed a `sys.error` including all the failed validation is displayed in the sbt console. Currently the following checks are verified:
  - If the settings values are compatible with the `conductrTargetVersion`
  - That one endpoint has either declared services or acls. Declaring both at the same endpoint is not allowed
- BundlePlugin: The `BundleKeys.endpoints` default has changed:
  - `enableAcls` == true: `Map(name.value -> Endpoint("http", 0, name.value))`
  - `enableAcls` == false: `Map(name.value -> Endpoint("http", 0, Set.empty[URI]))`
  - Meaning the endpoint key equals now the project name and no services or acls are exposed by default, i.e. the bundle can be located from the service locator but can't be accessed via the proxy
- LagomBundlePlugin: The calculation of the default endpoint for a Lagom service has changed:
  - `enableAcls` == true: Endpoints are collected in the acl format
  - `enableAcls` == false: Endpoints are collected in the services format (with `?preservePath`)
- LagomBundlePlugin: The default `BundleKeys.endpoints` for a Play service in Lagom is now `Map(name.value -> Endpoint("http", 0, Set(URI("http://:9000"))))`
- PlayBundlePlugin: The default `BundleKeys.endpoints` for a Play project is now `Map(name.value -> Endpoint("http", 0, Set(URI("http://:9000"))))`

Also a bunch of tests have been added to capture more corner cases.


markusjura:
Also a bug has been solved. In a lagom project the `sandbox` command has been executed per project. Now the command is only executed once. Fix: https://github.com/markusjura/sbt-conductr/blob/acl-and-service-support/src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala#L31

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -134,12 +121,49 @@ object BundlePlugin extends AutoPlugin {
...
+   * Checks whether the specified bundle settings are valid.
+   * If one or more validations are failing a sys.error is thrown
+   */
+  private def validateSettings: Def.Initialize[Task[Unit]] = Def.task {
+    // Checks that the project and global settings in the `BundlePlugin` are compatible with the `conductrTargetVersion`
+    def assertConductrVersionCompatibility: Seq[String] = {
+      val conductrVersion = conductrTargetVersion.value
+      def assertEnableAcls: Option[String] =
+        if (enableAcls.value && conductrVersion <= ConductrVersion.V1_1)
+          Some(s"'enableAcls' is set to 'true' and the 'conductrTargetVersion' is set to $conductrVersion. Acls are only supported since ConductR 1.2.")
```
huntc: Make a recommendation on what should be done.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -104,7 +105,7 @@ object ConductRPlugin extends AutoPlugin {
 
     val projectImageVersion = if (hasRpLicense.value) Some(LatestConductRVersion) else None
     val bundlePorts =
-      (BundleKeys.endpoints in Bundle).?.map(_.getOrElse(Map.empty)).all(filter).value
+      BundleKeys.endpoints.?.map(_.getOrElse(Map.empty)).all(filter).value
```
huntc: This looks fine, just curious about the consequences of the change. ReactiveMaps may be a good test for this.

src/main/scala/com/lightbend/conductr/sbt/LagomBundlePlugin.scala
```diff
@@ -38,8 +40,9 @@ object LagomPlayBundlePlugin extends AutoPlugin {
       BundleKeys.nrOfCpus := PlayBundleKeyDefaults.nrOfCpus,
       BundleKeys.memory := PlayBundleKeyDefaults.memory,
       BundleKeys.diskSpace := PlayBundleKeyDefaults.diskSpace,
-      conductrBundleLibVersion := Version.conductrBundleLib,
-      libraryDependencies += Library.lagomConductrBundleLib(LagomVersion.current, scalaBinaryVersion.value, conductrBundleLibVersion.value),
+      BundleKeys.endpoints := Map(name.value -> Endpoint("http", 0, Set(URI("http://:9000")))),
```
huntc: This should check the `enableAcls` flag to determine whether it should be expressed as an ACL or service URI.

src/main/scala/com/lightbend/conductr/sbt/LagomBundlePlugin.scala
```diff
@@ -201,8 +219,8 @@ object LagomBundlePlugin extends AutoPlugin {
    */
   private def collectEndpoints(config: Configuration): Def.Initialize[Task[Map[String, Endpoint]]] = Def.taskDyn {
     Def.task {
-      val manualEndpoints = (BundleKeys.endpoints in config).value
-      if (manualEndpoints != DefaultEndpoints)
+      val manualEndpoints = BundleKeys.endpoints.value
+      if (manualEndpoints != BundlePlugin.getDefaultEndpoints.value)
```
huntc: Needs the `config` scope.

huntc:
LGTM other than comments.

markusjura:
Changes related to the comments have been made: 
- `config` scope is now used at the appropriate places
- Validation messages have been improved
- For a Play project we set now the default endpoint either as acl or service endpoint depending on the flag `enableAcls`

huntc:
Cool - thanks. I've just renamed the title to WIP until you get the chance to try the plugins out with Reactive Maps.

markusjura:
PR has been tested with ConductR 1.1.2 against:
- activator-lagom-java
- activator-lagom-java-chirper
- ReactiveMaps

All tests have succeed. However, for ReactiveMaps it was necessary to disable the https://github.com/sbt/sbt-bintray-bundle plugin. This plugin should get integrated / or should get updated before releasing sbt-conductr 2.0.0. I'll do this in a separate PR.

huntc:
Why did we have to disable the bintray plugin? Was it because if dependencies?

markusjura:
Yes, if enabled I get a compile error because the project is looking for a method in `SbtBundle` and `BundlePlugin`. We've renamed the class into `BundlePlugin`.

markusjura:
Looking at the code of https://github.com/sbt/sbt-bintray-bundle/blob/master/src/main/scala/com/typesafe/sbt/bintraybundle/BintrayBundle.scala I think best is to upgrade `sbt-bintray-bundle` to `BundlePlugin` and leave it as a separate sbt plugin.

PR #149, opened 2016-04-11
Public sandbox run and stop methods

markusjura:
Making `sandbox run` and `sandbox stop` public so that `conductr` can use these methods in their acceptance tests.


huntc:
LGTM

PR #150, opened 2016-04-18
Use 'web' as default endpoint name

markusjura:
Using the endpoint name `default` instead of resolving the name based on the project name. Reason is to make it easier using the corresponding environment variables `DEFAULT_BIND_IP`, `DEFAULT_BIND_PORT` - particular when using the envs in a configuration file.

This PR is related to https://github.com/typesafehub/conductr-lib/pull/102.


huntc:
This feels quite tightly coupled with the lib...

huntc:
Actually ignore that - it is no worse than the coupling we already have for WEB. Let's just stick with WEB though.

markusjura:
Changed the default endpoint name from `default` to `web`.

PR #156, opened 2016-05-24
Introduces the install and generateInstallationScript commands

### LINKED ISSUE #142: Bundle configuration succeeds without bundle-configuration directory
The `configuration:dist` command generates a bundle configuration even if no `bundle-configuration` directory has been created for the project. Instead this situation should result into an sbt error.

The current code only throws this sbt error if:
- The generated bundle.conf is empty AND
- The `bundle-configuration` directory exists

https://github.com/typesafehub/sbt-conductr/blob/master/src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala#L427

Not sure why we have the `bundle.conf` check at that place. Tests are showing that the generated bundle.conf is always filled (e.g. with the compatibility version). Therefore I suggest to remove the `bundle.conf` check and only check if `bundle-configuration` the directory exists.

huntc:
A new `install` command is provided that conveniently installs all of the bundles and their associated configurations (If present) to the local sandbox cluster. In addition, if the project is a Lagom one then the Cassandra configuration will be generated and loaded along with Cassandra.

A new `generateInstallationScript` command generates an installation script that is encouraged to be copied and used as the basis of a script that can manage a collection of bundles, in particular a collection that represents a system of micro services as is the case with Lagom.

Along the way, a new `waitForConductr` function is provided for public usage of the `ConductrBundle`. It is expected that ConductR's own acceptance tests can make use of this function in place of their custom logic.

The latest conductr-libs, now available from Maven, have also been included. The Lagom library is updated for milestone 2, and so that had a ripple effect on our tests which have therefore been modified.

Lastly, this PR fixes #142 by emitting a configuration file only when some configuration settings have been provided.

### DIFF
--- a/src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
+++ b/src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
@@ -65,7 +65,7 @@ object BundlePlugin extends AutoPlugin {
       bundleConf := getConfig(config, forAllSettings = true).value,
       executableScriptPath := (file((normalizedName in config).value) / "bin" / (executableScriptName in config).value).getPath,
       NativePackagerKeys.packageName := (normalizedName in config).value + "-v" + (compatibilityVersion in config).value,
-      NativePackagerKeys.dist := Def.taskDyn(createDist(config, (bundleType in config).value)).value,
+      NativePackagerKeys.dist := Def.taskDyn(createBundle(config, (bundleType in config).value)).value,
       NativePackagerKeys.stage := Def.taskDyn(stageBundle(config, (bundleType in config).value)).value,
       NativePackagerKeys.stagingDirectory := (target in config).value / "stage",
       target := projectTarget.value / "bundle"
@@ -78,7 +78,6 @@ object BundlePlugin extends AutoPlugin {
     inConfig(config)(Seq(
       bundleConf := getConfig(config, forAllSettings = false).value,
       checks := Seq.empty,
-      compatibilityVersion := (version in config).value.takeWhile(_ != '.'),
       executableScriptPath := (file((normalizedName in config).value) / "bin" / (executableScriptName in config).value).getPath,
       NativePackagerKeys.dist := createConfiguration(config).value,
       NativePackagerKeys.stage := stageConfiguration(config).value,
@@ -164,7 +163,7 @@ object BundlePlugin extends AutoPlugin {
     extracted.structure.data.definingScope(scoped.scope, scoped.key).flatMap(_.config.toOption.map(_.name))
   }
 
-  private def createDist(config: Configuration, bundleTypeConfig: Configuration): Def.Initialize[Task[File]] = Def.task {
+  private def createBundle(config: Configuration, bundleTypeConfig: Configuration): Def.Initialize[Task[File]] = Def.task {
     validateSettings(config).value
 
     val bundleTarget = (target in config).value
@@ -182,12 +181,7 @@ object BundlePlugin extends AutoPlugin {
     )
   }
 
-  /**
-   * Creates a bundle configuration in the specified config target directory
-   *
-   * @param config under which the bundle configuration is created
-   * @return the created bundle configuration file
-   */
+  // Creates a bundle configuration in the specified config target directory
   private def createConfiguration(config: Configuration): Def.Initialize[Task[File]] = Def.task {
     validateSettings(config).value
 
@@ -201,6 +195,40 @@ object BundlePlugin extends AutoPlugin {
       bundleMappings,
       f => streams.value.log.info(s"Bundle configuration has been created: $f")
     )
+  }
+
+  private def stageBundle(config: Configuration, bundleTypeConfig: Configuration): Def.Initialize[Task[File]] = Def.task {
+    validateSettings(config).value
+
+    val bundleTarget = (NativePackagerKeys.stagingDirectory in config).value / config.name
+    writeConfig(bundleTarget, (bundleConf in config).value)
+    val componentTarget = bundleTarget / (normalizedName in config).value
+    IO.copy((mappings in bundleTypeConfig).value.map(p => (p._1, componentTarget / p._2)))
+    componentTarget
+  }
+
+  private def stageConfiguration(config: Configuration): Def.Initialize[Task[File]] = Def.task {
+    validateSettings(config).value
+
+    val configurationTarget = (NativePackagerKeys.stagingDirectory in config).value / config.name
+    val generatedConf = (bundleConf in config).value
+    val srcDir = (sourceDirectory in config).value / (configurationName in config).value
+    if (generatedConf.isEmpty && !srcDir.exists())
+      sys.error(
+        s"""Directory $srcDir does not exist.
+            | Specify the desired configuration directory name
+            |  with the 'configurationName' setting given that it is not "default"""".stripMargin
+      )
+    IO.createDirectory(configurationTarget)
+    if (generatedConf.nonEmpty) writeConfig(configurationTarget, generatedConf)
+    IO.copyDirectory(srcDir, configurationTarget, overwrite = true, preserveLastModified = true)
+    configurationTarget
+  }
+
+  private def writeConfig(target: File, contents: String): File = {
+    val configFile = target / "bundle.conf"
+    IO.write(configFile, contents, Utf8)
+    configFile
   }
 
   private final val DefaultEndpointName = "web"
@@ -245,6 +273,7 @@ object BundlePlugin extends AutoPlugin {
    * Creates a zip file for a given target directory.
    * The name of the zip file contains a SHA256 digest of the target directory,
    * meaning based on the content of the target directory a (unique) digest is created.
+   *
    * @param archiveTarget The target for which the zip archive is created
    * @param archiveName The first name part of the zip archive. The second part is the SHA256 digest
    * @param bundleMappings included in the output
@@ -395,93 +424,80 @@ object BundlePlugin extends AutoPlugin {
   }
 
   private def getConfig(config: Configuration, forAllSettings: Boolean): Def.Initialize[Task[String]] = Def.task {
-    val checkComponents = (checksConfigName in config).value match {
-      case (value, configName) if (forAllSettings || configName.isDefined) && value.nonEmpty =>
-        val checkInitialDelayValue = (checkInitialDelay in config).value
-        val checkInitialDelayInSeconds =
-          if (checkInitialDelayValue.toMillis % 1000 > 0)
-            checkInitialDelayValue.toSeconds + 1 // always round up
-          else
-            checkInitialDelayValue.toSeconds
-        Seq(
-          value.map(uri => s""""$uri"""").mkString(
-            s"""components = {
-                |  ${(normalizedName in config).value}-status = {
-                |    description      = "Status check for the bundle component"
-                |    file-system-type = "universal"
-                |    start-command    = ["check", "--initial-delay", "$checkInitialDelayInSeconds", """.stripMargin,
-            ", ",
-            s"""]
-                |    endpoints        = {}
-                |  }
-                |}""".stripMargin
-          )
-        )
-      case _ =>
-        Seq.empty[String]
-    }
+    /*
+     * This test is about determining if a value has been provided at a specific level of configuration,
+     * as distinct from delegation. We don’t want to determine the value of a given setting, but whether
+     * it has been defined at an exact point on the config axis...
+     * In the case of forAllSettings then we don’t care - we’re just going to write the value out.
+     */
+    def valueSuppliedForConfig[T](valueAndConfigName: (T, Option[String])): Boolean =
+      forAllSettings || valueAndConfigName._2.isDefined
 
-    def formatValue[T](format: String, valueAndConfigName: (T, Option[String])): Seq[String] =
-      valueAndConfigName match {
-        case (value, configName) if forAllSettings || configName.isDefined => Seq(format.format(value))
-        case _                                                             => Seq.empty[String]
+    if (valueSuppliedForConfig((normalizedNameConfigName in config).value) ||
+      valueSuppliedForConfig((compatibilityVersionConfigName in config).value) ||
+      valueSuppliedForConfig((systemConfigName in config).value) ||
+      valueSuppliedForConfig((systemVersionConfigName in config).value) ||
+      valueSuppliedForConfig((nrOfCpusConfigName in config).value) ||
+      valueSuppliedForConfig((memoryConfigName in config).value) ||
+      valueSuppliedForConfig((diskSpaceConfigName in config).value) ||
+      valueSuppliedForConfig((rolesConfigName in config).value) ||
+      valueSuppliedForConfig((projectInfoConfigName in config).value) ||
+      valueSuppliedForConfig((bundleTypeConfigName in config).value) ||
+      valueSuppliedForConfig((startCommandConfigName in config).value) ||
+      valueSuppliedForConfig((endpointsConfigName in config).value)) {
+
+      val checkComponents = (checksConfigName in config).value match {
+        case (value, configName) if (forAllSettings || configName.isDefined) && value.nonEmpty =>
+          val checkInitialDelayValue = (checkInitialDelay in config).value
+          val checkInitialDelayInSeconds = Math.max(1, checkInitialDelayValue.toSeconds)
+          Seq(
+            value.map(uri => s""""$uri"""").mkString(
+              s"""components = {
+                  |  ${(normalizedName in config).value}-status = {
+                  |    description      = "Status check for the bundle component"
+                  |    file-system-type = "universal"
+                  |    start-command    = ["check", "--initial-delay", "$checkInitialDelayInSeconds", """.stripMargin,
+              ", ",
+              s"""]
+                  |    endpoints        = {}
+                  |  }
+                  |}""".stripMargin
+            )
+          )
+        case _ =>
+          Seq.empty[String]
       }
 
-    def toString[T](valueAndConfigName: (T, Option[String]), f: T => String): (String, Option[String]) =
-      f(valueAndConfigName._1) -> valueAndConfigName._2
+      def formatValue[T](format: String, valueAndConfigName: (T, Option[String])): Seq[String] =
+        if (valueSuppliedForConfig(valueAndConfigName))
+          Seq(format.format(valueAndConfigName._1))
+        else
+          Seq.empty[String]
 
-    val declarations =
-      Seq(s"""version              = "${bundleConfVersion.value}"""") ++
-        formatValue("""name                 = "%s"""", (normalizedNameConfigName in config).value) ++
-        formatValue("""compatibilityVersion = "%s"""", (compatibilityVersionConfigName in config).value) ++
-        formatValue("""system               = "%s"""", (systemConfigName in config).value) ++
-        formatValue("""systemVersion        = "%s"""", (systemVersionConfigName in config).value) ++
-        formatValue("nrOfCpus             = %s", (nrOfCpusConfigName in config).value) ++
-        formatValue("memory               = %s", toString((memoryConfigName in config).value, (v: Bytes) => v.underlying.toString)) ++
-        formatValue("diskSpace            = %s", toString((diskSpaceConfigName in config).value, (v: Bytes) => v.underlying.toString)) ++
-        formatValue(s"roles                = %s", toString((rolesConfigName in config).value, (v: Set[String]) => formatSeq(v))) ++
-        Seq("components = {", s"  ${(normalizedName in config).value} = {") ++
-        formatValue(s"""    description      = "%s"""", toString((projectInfoConfigName in config).value, (v: ModuleInfo) => v.description)) ++
-        formatValue(s"""    file-system-type = "%s"""", (bundleTypeConfigName in config).value) ++
-        formatValue(s"""    start-command    = %s""", toString((startCommandConfigName in config).value, (v: Seq[String]) => formatSeq(v))) ++
-        formatValue(s"""    endpoints = %s""", toString((endpointsConfigName in config).value, (v: Map[String, Endpoint]) => formatEndpoints(bundleConfVersion.value, v))) ++
-        Seq("  }", "}") ++
-        checkComponents
+      def toString[T](valueAndConfigName: (T, Option[String]), f: T => String): (String, Option[String]) =
+        f(valueAndConfigName._1) -> valueAndConfigName._2
 
-    declarations.mkString("\n")
-  }
+      val declarations =
+        Seq(s"""version              = "${bundleConfVersion.value}"""") ++
+          formatValue("""name                 = "%s"""", (normalizedNameConfigName in config).value) ++
+          formatValue("""compatibilityVersion = "%s"""", (compatibilityVersionConfigName in config).value) ++
+          formatValue("""system               = "%s"""", (systemConfigName in config).value) ++
+          formatValue("""systemVersion        = "%s"""", (systemVersionConfigName in config).value) ++
+          formatValue("nrOfCpus             = %s", (nrOfCpusConfigName in config).value) ++
+          formatValue("memory               = %s", toString((memoryConfigName in config).value, (v: Bytes) => v.underlying.toString)) ++
+          formatValue("diskSpace            = %s", toString((diskSpaceConfigName in config).value, (v: Bytes) => v.underlying.toString)) ++
+          formatValue(s"roles                = %s", toString((rolesConfigName in config).value, (v: Set[String]) => formatSeq(v))) ++
+          Seq("components = {", s"  ${(normalizedName in config).value} = {") ++
+          formatValue(s"""    description      = "%s"""", toString((projectInfoConfigName in config).value, (v: ModuleInfo) => v.description)) ++
+          formatValue(s"""    file-system-type = "%s"""", (bundleTypeConfigName in config).value) ++
+          formatValue(s"""    start-command    = %s""", toString((startCommandConfigName in config).value, (v: Seq[String]) => formatSeq(v))) ++
+          formatValue(s"""    endpoints = %s""", toString((endpointsConfigName in config).value, (v: Map[String, Endpoint]) => formatEndpoints(bundleConfVersion.value, v))) ++
+          Seq("  }", "}") ++
+          checkComponents
 
-  private def stageBundle(config: Configuration, bundleTypeConfig: Configuration): Def.Initialize[Task[File]] = Def.task {
-    validateSettings(config).value
-
-    val bundleTarget = (NativePackagerKeys.stagingDirectory in config).value / config.name
-    writeConfig(bundleTarget, (bundleConf in config).value)
-    val componentTarget = bundleTarget / (normalizedName in config).value
-    IO.copy((mappings in bundleTypeConfig).value.map(p => (p._1, componentTarget / p._2)))
-    componentTarget
-  }
-
-  private def stageConfiguration(config: Configuration): Def.Initialize[Task[File]] = Def.task {
-    validateSettings(config).value
-
-    val configurationTarget = (NativePackagerKeys.stagingDirectory in config).value / config.name
-    val generatedConf = (bundleConf in config).value
-    val srcDir = (sourceDirectory in config).value / (configurationName in config).value
-    if (generatedConf.isEmpty && !srcDir.exists())
-      sys.error(
-        s"""Directory $srcDir does not exist.
-            | Specify the desired configuration directory name
-            |  with the 'configurationName' setting given that it is not "default"""".stripMargin
-      )
-    IO.createDirectory(configurationTarget)
-    if (generatedConf.nonEmpty) writeConfig(configurationTarget, generatedConf)
-    IO.copyDirectory(srcDir, configurationTarget, overwrite = true, preserveLastModified = true)
-    configurationTarget
-  }
-
-  private def writeConfig(target: File, contents: String): File = {
-    val configFile = target / "bundle.conf"
-    IO.write(configFile, contents, Utf8)
-    configFile
+      declarations.mkString("\n")
+    } else {
+      ""
+    }
   }
 }
--- a/src/main/scala/com/lightbend/conductr/sbt/ConductRImport.scala
+++ b/src/main/scala/com/lightbend/conductr/sbt/ConductRImport.scala
@@ -4,32 +4,74 @@
 
 package com.lightbend.conductr.sbt
 
+import java.nio.file.Path
+
 import sbt._
 
 object ConductrImport {
 
+  object InstallationData {
+    /**
+     * Flatten our Either type to a string.
+     */
+    def nameOrPath(e: Either[String, Path]): String =
+      e match {
+        case Left(v)  => v
+        case Right(v) => v.toString
+      }
+  }
+
+  /**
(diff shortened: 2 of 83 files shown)


src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -203,6 +197,40 @@ object BundlePlugin extends AutoPlugin {
...
+    IO.copyDirectory(srcDir, configurationTarget, overwrite = true, preserveLastModified = true)
+    configurationTarget
+  }
+
+  private def writeConfig(target: File, contents: String): File = {
+    val configFile = target / "bundle.conf"
+    IO.write(configFile, contents, Utf8)
+    configFile
+  }
+
```
huntc: These functions were at the bottom of the source file and separated from their `dist` relations. I felt that it was nicer to have them back here and accordingly grouped.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -395,93 +424,81 @@ object BundlePlugin extends AutoPlugin {
   }
 
   private def getConfig(config: Configuration, forAllSettings: Boolean): Def.Initialize[Task[String]] = Def.task {
```
huntc: The changes to `getConfig` are with respect to testing whether a configuration file actually does need to be generated, as it was always being generated before.

src/main/scala/com/lightbend/conductr/sbt/ConductRImport.scala
```diff
@@ -10,26 +10,43 @@ object ConductrImport {
...
+
+    val generateInstallationScript = TaskKey[File](
+      "conductr-generate-installation-script",
+      "Produces the installation script to install the entire build's bundles"
+    )
+
+    val install = TaskKey[File](
+      "install",
+      "Runs an installation script from generateInstallationScript that will install an entire build's bundles"
+    )
```
huntc: These new keys represent the new functionality offered by this PR.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -27,7 +33,7 @@ object ConductrPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   override def globalSettings: Seq[Setting[_]] =
-    super.globalSettings ++ List(
```
huntc: Settings don't need to declare their super settings as what is declared here will be appended anyhow.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +51,125 @@ object ConductrPlugin extends AutoPlugin {
...
 
+      discoveredDist <<= (dist in Bundle).storeAs(discoveredDist).triggeredBy(dist in Bundle),
+      discoveredConfigDist <<= (dist in BundleConfiguration).storeAs(discoveredConfigDist).triggeredBy(dist in BundleConfiguration)
+    )
+
+  override def buildSettings: Seq[Setting[_]] =
+    List(
       sandbox := sandboxTask.value.evaluated,
       sandboxRunTaskInternal := sandboxRunTask(ScopeFilter(inAnyProject, inAnyConfiguration)).value,
       conduct := conductTask.value.evaluated,
```
huntc: The above 3 are now `buildSettings` - makes sense as they apply to the build as a whole.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +51,125 @@ object ConductrPlugin extends AutoPlugin {
...
+      install := installTask().value
     )
 
-  private final val LatestConductrVersion = "1.1.3"
+  private final val LatestConductrVersion = "1.1.5"
   private final val LatestConductrDocVersion = LatestConductrVersion.dropRight(1) :+ "x" // 1.0.0 to 1.0.x
 
   private final val TypesafePropertiesName = "typesafe.properties"
   private final val SandboxRunArgsAttrKey = AttributeKey[SandboxRunArgs]("conductr-sandbox-run-args")
   private final val sandboxRunTaskInternal = taskKey[Unit]("Internal Helper to call sandbox run task.")
```
huntc: What follows represents the bulk of the new functionality.

src/main/scala/com/lightbend/conductr/sbt/LagomConductRPlugin.scala
```diff
@@ -0,0 +1,41 @@
+package com.lightbend.conductr.sbt
```
huntc: This is a new plugin that is enabled when Lagom and the ConductR plugins are enabled. The plugin writes out Cassandra installation script contents.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -395,93 +424,81 @@ object BundlePlugin extends AutoPlugin {
...
-        Seq.empty[String]
-    }
-
-    def formatValue[T](format: String, valueAndConfigName: (T, Option[String])): Seq[String] =
+    def hasValue[T](valueAndConfigName: (T, Option[String])): Boolean =
       valueAndConfigName match {
-        case (value, configName) if forAllSettings || configName.isDefined => Seq(format.format(value))
-        case _                                                             => Seq.empty[String]
+        case (value, configName) if forAllSettings || configName.isDefined => true
+        case _                                                             => false
```
dotta: I'd refactor this method as:

```
def hasValue[T](valueAndConfigName: (T, Option[String])): Boolean =
  forAllSettings || valueAndConfigName._2.isDefined
```

I should also say that to me is not clear what the implementation does. Based on the method's name (`hasValue`), I'd assume that the implementation should answer the question "does this pair have a value?". But, when I look at the implementation, it doesn't even look at the value, so I'm left wondering what this method is actually for. Sorry for the rant :)

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -395,93 +424,81 @@ object BundlePlugin extends AutoPlugin {
...
+      hasValue((endpointsConfigName in config).value)) {
+
+      val checkComponents = (checksConfigName in config).value match {
+        case (value, configName) if (forAllSettings || configName.isDefined) && value.nonEmpty =>
+          val checkInitialDelayValue = (checkInitialDelay in config).value
+          val checkInitialDelayInSeconds =
+            if (checkInitialDelayValue.toMillis % 1000 > 0)
+              checkInitialDelayValue.toSeconds + 1 // always round up
+            else
+              checkInitialDelayValue.toSeconds
```
dotta: Maybe the above if-then-else be changed to something like `math.floor(checkInitialDelayValue.toSeconds)`?

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -395,93 +424,81 @@ object BundlePlugin extends AutoPlugin {
...
+          Seq(
+            value.map(uri => s""""$uri"""").mkString(
```
dotta: Why do you need to `map`?  Calling `mkString` directly should be enough I think

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
dotta: By making these build tasks, won't the task be executed on all project aggregates? Just pointing this out as I remember noticing this in the Lagom plugin, and in fact we have code to prevent aggregation (see https://github.com/lagom/lagom/blob/master/dev/sbt-plugin/src/main/scala/com/lightbend/lagom/sbt/LagomPlugin.scala#L311-L317)

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +51,125 @@ object ConductrPlugin extends AutoPlugin {
...
+      tasks match {
+        case Nil     => Def.task(a)
+        case x :: xs => Def.taskDyn(fold(a :+ x.value)(xs))
+      }
+
+    fold(List.empty)(bundleInstallationScriptsTasks)
+  }
+
+  private def generateInstallationScriptTask(): Def.Initialize[Task[File]] = Def.task {
+    val installationScript = (target in LocalRootProject).value / "install.sh"
```
dotta: I guess we don't support Windows (just pointing it out in case it wasn't intentional)

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -170,6 +276,37 @@ object ConductrPlugin extends AutoPlugin {
   def sandboxStop(): Unit =
     Process(Seq("sandbox", "stop")).!
 
+  /**
+   * A convenience function that waits on ConductR to become available.
+   * @param duration the max time to wait.
+   */
+  def waitForConductr(implicit duration: FiniteDuration): Unit = {
+    println("Waiting for ConductR to start...")
```
dotta: It would be nice if we could provide users with feedback that something is happening in the background, like printing a dot every 500 milliseconds (we do exactly that when starting/stopping stuff in the lagom plugin, see https://github.com/lagom/lagom/blob/master/dev/sbt-plugin/src/main/scala/com/lightbend/lagom/sbt/run/Servers.scala#L137)

src/main/scala/com/lightbend/conductr/sbt/LagomConductRPlugin.scala
```diff
@@ -0,0 +1,41 @@
...
+  import SbtNativePackager.autoImport._
+  import NativePackagerKeys._
+
+  override def requires = ConductrPlugin && LagomBundlePlugin
+
+  override def trigger = allRequirements
+
+  override def buildSettings: Seq[Setting[_]] =
+    List(
+      installationScriptContents ++= installationScriptContentsTask.value
```
dotta: The same comment I made [here](https://github.com/typesafehub/sbt-conductr/pull/156/files#r64393921) may apply

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: Thanks - yeah, the function's changes makes this doable.

huntc: I've also renamed the function to `valueSuppliedForConfig`. Hopefully that makes more sense?

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: Thanks. Now:

```
          val checkInitialDelayInSeconds = Math.max(1, checkInitialDelayValue.toSeconds)
```

I just want to ensure that 0 doesn't get through.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: It is surrounding the value with double quotes. Changed in order to make this clearer:

```
            value.map(uri => s"\"$uri\"").mkString(
```

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Thanks, I needed to address this. Now we have:

```
      Keys.aggregate in conduct := false,
      Keys.aggregate in generateInstallationScript := false,
      Keys.aggregate in install := false,
      Keys.aggregate in sandbox := false,
```

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: You raise a good point. The recent changes we had made to sbt-conductr in support of providing a translation to CLI commands has actually broken this requirement. For example, we're using `awk`:

```
              (Process("conduct info") #| Process(Seq("awk", "{print $2}")) #| Process(Seq("awk", "{if(NR>1)print}"))).lines(NoProcessLogging).toSet
```

I'll raise a ticket for this separately.

huntc: On this topic, I wonder if it is now reasonable given bash's availability for Windows 10, whether we have to support Windows specifically...

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Done - thanks

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
dotta: Much better :) Though, if `forAllSettings` is true, the config name can be missing. Out of curiosity, why is that? :)

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
dotta: 👍

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
dotta: Thx, didn't notice the extra quote

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
dotta: Might be, not sure what's the market share of Windows 10.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
```diff
@@ -395,93 +424,74 @@ object BundlePlugin extends AutoPlugin {
...
+      valueSuppliedForConfig((rolesConfigName in config).value) ||
+      valueSuppliedForConfig((projectInfoConfigName in config).value) ||
+      valueSuppliedForConfig((bundleTypeConfigName in config).value) ||
+      valueSuppliedForConfig((startCommandConfigName in config).value) ||
+      valueSuppliedForConfig((endpointsConfigName in config).value)) {
+
+      val checkComponents = (checksConfigName in config).value match {
+        case (value, configName) if (forAllSettings || configName.isDefined) && value.nonEmpty =>
+          val checkInitialDelayValue = (checkInitialDelay in config).value
+          val checkInitialDelayInSeconds = Math.max(1, checkInitialDelayValue.toSeconds)
```
dotta: Is the value of `checkInitialDelayValue.toSeconds` round up? Just asking as otherwise there is a small difference with the former implementation

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: This test is about determining if a value has been provided at a specific level of configuration, as distinct from delegation. We don’t want to determine the value of a given setting, but whether it has been defined at an exact point on the config axis…

For example:

lazy val BackendRegion = config("backend-region").extend(Bundle)
BundlePlugin.bundleSettings(BackendRegion)
inConfig(BackendRegion)(Seq(
  normalizedName := "reactive-maps-backend-region",
  BundleKeys.configurationName := (normalizedName in BackendRegion).value,
  ...
))

That test will return true for normalizedName only if it has been defined at the BackendRegion config level. Thus we know that we need to write the value out in our bundle.

In the case of forAllSettings then we don’t care - we’re going to write the value out.

Does this make sense?

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: I’m eradicating the Unix specific stuff now.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
dotta: Got it. I understand what's doing now, and I'd suggest adding a comment as it is not straightforward :)

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
dotta: I like "eradicating"... looking forward to have a chance to use the verb myself ;-)

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: Looking at it again, I just want to ensure that it isn’t zero.

src/main/scala/com/lightbend/conductr/sbt/BundlePlugin.scala
huntc: Comment now added. Thanks.

dotta:
lgtm - it'd be good if also @markusjura could review this PR

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -304,10 +471,14 @@ object ConductrPlugin extends AutoPlugin {
...
+          val bundleNames =
             withProcessHandling {
-              (Process("conduct info") #| Process(Seq("awk", "{print $2}")) #| Process(Seq("awk", "{if(NR>1)print}"))).lines(NoProcessLogging).toSet
-            }(Set.empty)
+              "conduct info"
+                .lines_!(NoProcessLogging)
+                .slice(1, 11) //                         No more than this number of lines please, and no header...
+                .flatMap(_.split("\\s+").slice(1, 2)) // Just the second column i.e. the name
+                .toSet
+            }(Set.empty[String])
```
huntc: This change permits tab completion for Windows (it was Unix only before)

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -170,6 +312,31 @@ object ConductrPlugin extends AutoPlugin {
...
+  def waitForConductr(implicit duration: FiniteDuration): Unit = {
+    print("Waiting for ConductR to start")
+
+    val deadline = duration.fromNow
+
+    @tailrec
+    def loop(deadline: Deadline): Unit = {
+      print(".")
+      if (deadline.isOverdue())
+        throw new TimeoutException(s"ConductR has not been started within ${duration.toSeconds} seconds!")
```
dotta: Not sure how the exception will be displayed to the user (doesn't one have to type `last`?), but to me it seems a log message is a nicer way of telling the user what went wrong (that's what I'm doing in Lagom, see [here](https://github.com/lagom/lagom/blob/master/dev/sbt-plugin/src/main/scala/com/lightbend/lagom/sbt/run/Servers.scala#L158) for an example)

dotta: I see you are also using `sys.error` in some places, and I'm wondering if `log.error` isn't better.

huntc: The message is logged as an error at the console... which I think is better than colouring yourself. ;-)

huntc: `sys.error` will exit - which is what we want.

dotta: Alright, thx for the info. Though, exiting sbt (?) seems a bit drastic

dotta: Ok, `sys.error` throws an exception, so all good.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
 
-  private final val LatestConductrVersion = "1.1.3"
+  private final val LatestConductrVersion = "1.1.5"
   private final val LatestConductrDocVersion = LatestConductrVersion.dropRight(1) :+ "x" // 1.0.0 to 1.0.x
 
   private final val TypesafePropertiesName = "typesafe.properties"
   private final val SandboxRunArgsAttrKey = AttributeKey[SandboxRunArgs]("conductr-sandbox-run-args")
   private final val sandboxRunTaskInternal = taskKey[Unit]("Internal Helper to call sandbox run task.")
 
+  private def installationDataTask: Def.Initialize[Task[Seq[InstallationData]]] = Def.taskDyn {
```
markusjura: Is a `taskDyn` necessary here or can we also use a `Def.task`?

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+        val bundleConfiguration = project.configurations.find(c => c.name == Bundle.name)
+        val bundleConfigurationConfiguration = project.configurations.find(c => c.name == BundleConfiguration.name)
+
+        (bundleConfiguration, bundleConfigurationConfiguration) match {
+          case (Some(b), Some(bc)) =>
+            Def.taskDyn {
+              val bundleName = (normalizedName in b in projectRef).value
+              val bundleFile = (dist in b in projectRef).value.toPath
+              val bundleConfigContents = (BundleKeys.bundleConf in bc in projectRef).value
+              val srcDir = (sourceDirectory in bc in projectRef).value / (BundleKeys.configurationName in bc in projectRef).value
```
markusjura: Can we rename `srcDir` to `configurationDir` or `bundleConfigSrcDir`? It is the directory of the bundle configuration.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
   private final val sandboxRunTaskInternal = taskKey[Unit]("Internal Helper to call sandbox run task.")
 
+  private def installationDataTask: Def.Initialize[Task[Seq[InstallationData]]] = Def.taskDyn {
+    val rootProjectStructure = Project.extract(state.value).structure
+    val allProjects = rootProjectStructure.allProjects
+    val bundleInstallationScriptsTasks =
+      allProjects.map { project =>
+        val projectRef = ProjectRef(rootProjectStructure.root, project.id)
+
+        val bundleConfiguration = project.configurations.find(c => c.name == Bundle.name)
```
markusjura: Can we rename `bundleConfiguration` and `bundleConfigurationConfiguration`. I've found it confusing and didn't understand in the beginning that the first "configuration" is the sbt project configuration.

What about: `BundleConfig` and `BundleConfConfig`

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+    IO.write(
+      installationScript,
+      installationScriptContents.mkString(
+        s"""#!/usr/bin/env bash
+            |cd "$$( dirname "$${BASH_SOURCE[0]}" )"
+            |""".stripMargin,
+        "",
+        """
+          |echo 'Your system is deployed. Running "conduct info" to observe the cluster.'
+          |echo 'Type just "conduct" for help.'
```
markusjura: Not sure if we need the line "echo 'Type just "conduct" for help.'". At that point in time the user should familiar with the conduct command. Also he might want to use `conduct logs`, depending on the situation.

If we stick with the line I would rename it to: `echo 'Type "conduct" for help.'`

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+          |conduct info
+          |""".stripMargin
+      ),
+      Utf8
+    )
+    println("\nYour system's ConductR installation script: " + installationScript)
```
markusjura: What about: "\nThe ConductR installation script has been successfully created at: "

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+        println("Restarting ConductR to ensure a clean state...")
+        "docker exec cond-0 rm /opt/conductr/conf/seed-nodes".! // Allow the first node to form the new cluster
+        for (n <- 0 until nrOfContainers) s"docker restart -t 0 cond-$n".!
+      } else {
+        throw new IllegalStateException("Please first start the sandbox using 'sandbox run'.")
+      }
+    }(sys.error("There was a problem re/starting the sandbox."))
+
+    withProcessHandling {
+      waitForConductr(20.seconds)
```
markusjura: Should we make this duration a settingskey and therefore customizable by the user?

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+
+      installationData.value.foreach {
+        case InstallationData(bundleName, bundle, bundleConfigFile) =>
+          println(s"Deploying $bundleName...")
+          val bundleArg = InstallationData.nameOrPath(bundle)
+          val bundleConfigArg = bundleConfigFile.fold("")(_ + " ")
```
markusjura: `val bundleConfigArg = bundleConfigFile.mkString(" ")` should work as well.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+          s"conduct load $bundleArg $bundleConfigArg--long-ids -q".lines_!.headOption match {
+            case Some(bundleId) =>
+              if (s"conduct run $bundleId --no-wait -q".! != 0)
+                throw new IllegalStateException("The bundle could not run")
+            case None =>
+              throw new IllegalStateException("The bundle could not be loaded")
```
markusjura: Add the `bundleArg` to the error message: `throw new IllegalStateException(s"The bundle $bundleArg could not be loaded")`

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+
+      installationData.value.foreach {
+        case InstallationData(bundleName, bundle, bundleConfigFile) =>
+          println(s"Deploying $bundleName...")
+          val bundleArg = InstallationData.nameOrPath(bundle)
+          val bundleConfigArg = bundleConfigFile.fold("")(_ + " ")
+          s"conduct load $bundleArg $bundleConfigArg--long-ids -q".lines_!.headOption match {
+            case Some(bundleId) =>
+              if (s"conduct run $bundleId --no-wait -q".! != 0)
+                throw new IllegalStateException("The bundle could not run")
```
markusjura: Add the `bundleArg` to the error message: `throw new IllegalStateException(s"The bundle $bundleArg could not run")`

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -45,26 +47,165 @@ object ConductrPlugin extends AutoPlugin {
...
+            case None =>
+              throw new IllegalStateException("The bundle could not be loaded")
+          }
+      }
+
+    }(sys.error("There was a problem installing the project to ConductR."))
+
+    println
+    println("""Your system is deployed. Running "conduct info" to observe the cluster.""")
+    println("""Type just "conduct" for help.""")
```
markusjura: Not sure if we need the line "echo 'Type just "conduct" for help.'". At that point in time the user should familiar with the conduct command. Also he might want to use `conduct logs`, depending on the situation.

If we stick with the line I would rename it to: `echo 'Type "conduct" for help.'`

src/main/scala/com/lightbend/conductr/sbt/LagomConductRPlugin.scala
markusjura: Nice!

markusjura:
Overall really great improvements and nice implementation. Well done. Just added a couple of minor comments (typos, coding conventions, naming). Nothing drastic..

markusjura:
LGTM

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Yes, it is necessary as there are sub tasks that are resolved dynamically.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Renamed to `bundleConfigSrcDir`

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: I understand, however it is consistent with naming elsewhere. The `Configuration` as a suffix relates to sbt's `Configuration`.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: I'll remove the line.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Done

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: I've now provided a default param to `waitForConductr` of 20 seconds and removed it from here. We could make it a setting, although I'd prefer that we control it - we should know how long it should take.

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Nice catch!

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Done

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Done

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
huntc: Removed line

PR #157, opened 2016-06-03
"conduct load <tab>" no longer worked

huntc:
The last PR moved the "conduct" command to the build scope. This was an oversight in that it requires project settings. The command's scope has been restored.


edwardcallahan:
LGTM

PR #160, opened 2016-06-28
Using double quotes in docker command

markusjura:
With docker 1.12 the single quote in the `docker ps` command was interpreted as two single quotes by docker and therefore resulted into this error:

```
Error response from daemon: Invalid filter ''name'
```

This PR fixes the error by using double quotes instead of single quotes.


huntc:
LGTM

markusjura:
jenkins

PR #162, opened 2016-06-30
Correct semver handling for image

huntc:
ConductR images support semantic versioning, which also means the addition of a `-rc.1` style of suffix. Prior to this PR, declaring `sandbox run 1.1.8-rc.1` did not parse. It now does.


markusjura:
LGTM

PR #164, opened 2016-07-11
Latest conductr-lib

huntc:
The latest conductr-lib fixes a problem where the service locator would cache the fact that an address could not be found.


markusjura:
LGTM

huntc:
Tested running conductr and conductr agent via stage. Expanded zips:

screen shot 2016-07-11 at 9 35 01 pm

screen shot 2016-07-11 at 9 35 09 pm

PR #165, opened 2016-07-15
Set default nrOfCpus for Play and Lagom to 0.1

markusjura:
The `nrOfCpus` paramter is considered during the scheduling of a bundle. If a resource offer has less CPU then the bundle has been specified then the offer will be declined.

The `nrOfCpus` paramter is not considered while the bundle is running. Each bundle running on a node has access to all available CPUs on the node and therefore sharing the CPUs on the node.

During bundle scaling Mesos considers the specified CPU allocation of running bundles when accepting or declining an offer.

**Example**
- Node: 8 CPUs
- Bundle A: 4 CPUs
- Bundle B: 6 CPUs

In Mesos, once bundle A is running, Mesos will send a new resource offer with a CPU resource of 4. The original 8 CPUs are getting reduced by the bundle A, when accepting the offer. Therefore, bundle B will decline the resource offer because it needs 6 CPUs to run.

In standalone mode bundle B is accepted, because the simple resource provider of ConductR does not consider the CPU consumtion of running bundles. It only checks if the node has more CPUs available as the bundle has specified.

**What does this PR is changing?**
This PR changes the default `nrOfCpus` for Lagom and Play applications from 1.0 to 0.1 to respect the fact that Mesos considers the specified CPU allocating of running bundles when accepting or declining an offer.


src/main/scala/com/lightbend/conductr/sbt/package.scala
```diff
@@ -92,7 +92,7 @@ package object sbt {
   object PlayBundleKeyDefaults {
     import com.lightbend.conductr.sbt.BundleImport.ByteConversions._
 
-    val nrOfCpus = 1.0
+    val nrOfCpus = 0.1
```
markusjura: This is the main change of the PR. The rest is just updating documentation and test data.

huntc:
LGTM

PR #168, opened 2016-08-30
Use Lagom major version only when resolving ConductR Bundle Lib for Lagom

fsat:
This is due to the fact that Lagom will support semantic versioning, and hence all releases within the same major versions will be binary compatible.

Also upgrade to ConductR Bundle Lib 1.4.8 since ConductR Bundle Lib 1.4.8 has been made with Lagom's use of semantic versioning.

### DIFF
--- a/src/main/scala/com/lightbend/conductr/sbt/package.scala
+++ b/src/main/scala/com/lightbend/conductr/sbt/package.scala
@@ -64,16 +64,19 @@ package object sbt {
 
   object Library {
     def playConductrBundleLib(playVersion: String, scalaVersion: String, conductrLibVersion: String) =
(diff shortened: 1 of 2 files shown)


fsat:
Marked as wip so I can link this through Trello.

Still need to release a new version of ConductR Lib, waiting for access to be granted: https://issues.sonatype.org/browse/OSSRH-24675

Build will only pass after new ConductR Lib has been released due to name change in the ConductR Lib lagom artefact.

fsat:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;dDNzqJqe&#x2F;321-reorganise-conductr-lib-in-order-to-leverage-lagom-s-semantic-versioning"> Reorganise conductr-lib in order to leverage Lagom&#39;s semantic versioning</a>

huntc:
LGTM

PR #174, opened 2016-10-23
Fix sandbox run feature parser

markusjura:
The `sandbox run` parser supports now the following `--feature` format:

```
sandbox run 1.1.9 --feature monitoring
sandbox run 1.1.9 --feature monitoring v2
sandbox run 1.1.9 --feature monitoring 2.1.0-RC2
```

Allowed format: `--feature <name> <optional_arg>`

`conductr-cli` will then check if the given feature is valid.

### DIFF
--- a/src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
+++ b/src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
@@ -354,8 +354,6 @@ object ConductrPlugin extends AutoPlugin {
     object Sandbox {
       import ArgumentConverters._
 
-      val availableFeatures = Set("visualization", "logging", "monitoring")
-
       // Sandbox parser
       lazy val subtask: Def.Initialize[State => Parser[SandboxSubtask]] = Def.value {
         case _ =>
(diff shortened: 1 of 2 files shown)


src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -430,11 +428,8 @@ object ConductrPlugin extends AutoPlugin {
         Space ~> (token(Flags.port) | hideAutoCompletion("-p")) ~> numberWithText("<port>").map(PortArg(_))
 
       def feature: Parser[FeatureArg] =
-        Space ~> (token(Flags.feature) | hideAutoCompletion("-f")) ~> (featureExamples ~ nonArgStringWithText("<feature_arg>").*)
-          .map { case (feature, args) => FeatureArg(feature +: args) }
-
-      def featureExamples: Parser[String] =
-        Space ~> token(StringBasic.examples(availableFeatures))
+        Space ~> (token(Flags.feature) | hideAutoCompletion("-f")) ~> nonArgStringWithText(s"Format: ${Flags.feature} <name> <optional_arg>").+
+          .map(args => FeatureArg(args))
```
pvlugter: `featureExamples ~ args.*` and `args.+` are basically equivalent, and they both give optional additional args, but the first one still gives auto-completion on the feature names.

So before it was possible to:

```
sbt root ❯ sandbox run 1.1.10 -f <tab>
logging         monitoring      visualization
sbt root ❯ sandbox run 1.1.10 -f m<tab>
sbt root ❯ sandbox run 1.1.10 -f monitoring
sbt root ❯ sandbox run 1.1.10 -f monitoring <tab>
--conductr-role      --env                --feature            --image              --log-level
--nr-of-containers   --port               <conductr_version>   <feature_arg>
sbt root ❯ sandbox run 1.1.10 -f monitoring 2.1.0
```

But now it's:

```
sbt root ❯ sandbox run 1.1.10 -f <tab>
Format: --feature <name> <optional_arg>
sbt root ❯ sandbox run 1.1.10 -f m<tab>
Format: --feature <name> <optional_arg>
sbt root ❯ sandbox run 1.1.10 -f monitoring <tab>
--conductr-role                           --env
--feature                                 --image
--log-level                               --nr-of-containers
--port                                    <conductr_version>
Format: --feature <name> <optional_arg>
sbt root ❯ sandbox run 1.1.10 -f monitoring 2.1.0
```

where `monitoring` needs to be written in full each time, and there's no hint for which features are available.

I think the parser before this change was more useful.

PR #177, opened 2016-11-01
Windows file paths for start command

huntc:
The start command must always be expressed in Unix file path terms, nothing else (including Windows).


markusjura:
Manual test on Windows was successful.

PR #180, opened 2016-11-21
Update sandbox commands

markusjura:
- Removing the sandbox init command
- Adding the sandbox version command


markusjura:
Mostly. The parsing code is python is much nicer than in Scala, though ;-)

PR #182, opened 2016-11-24
Validates resident memory

huntc:
This memory checking verification should help users avoid specifying resident memory values that are too low at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,13 +255,13 @@ javaOptions in Universal := Seq(
 )
 
 BundleKeys.nrOfCpus := 0.1
-BundleKeys.memory := 128.MiB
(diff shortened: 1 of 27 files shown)


huntc:
Sample output:

```
[error]   - The bundle's resident 'memory' setting of 64 MiB is too low. Either increase it or lower the bundle's `minMemoryCheckValue` setting of 384 MiB. Note that a bundle's resident memory is not Java heap memory. Also note that exceeding this memory at runtime can cause your process to be killed unexpectedly. Tip: Use the Unix `top` command to determine this value by observing the `RES` column for your process and then rounding up to the nearest 10 MiB.
```

PR #190, opened 2017-02-05
Introduce --nr-of-instances and diminish Docker usage

huntc:
This commit introduces the new `--nr-of-instances` arg, and Diminishes the use of Docker from our tests.


src/sbt-test/sbt-conductr/sandbox-conductr-roles/build.sbt
```diff
@@ -21,25 +21,24 @@ BundleKeys.roles := Set("bundle-role-1", "bundle-role-2")
 
 val checkConductrRolesByBundle = taskKey[Unit]("Check that the bundle roles are used if no sandbox roles are specified.")
 checkConductrRolesByBundle := {
-  for (i <- 0 to 2) {
-    val content = s"docker inspect --format='{{.Config.Env}}' cond-$i".!!
-    val expectedContent = "CONDUCTR_ROLES=bundle-role-1,bundle-role-2"
-    content should not include(expectedContent)
+  val psOutput = s"ps ax".lines_!.toList
```
longshorej: Would it make sense to add bundle roles to `sandbox ps` and using it instead of `ps aux`, thus making this test a bit more portable?

src/main/scala/com/lightbend/conductr/sbt/ConductRPlugin.scala
```diff
@@ -292,6 +294,10 @@ object ConductrPlugin extends AutoPlugin {
         conductrImageVersion.toSeq ++
         conductrImage.withFlag(Flags.image) ++
         nrOfContainers.withFlag(Flags.nrOfContainers) ++
+        nrOfInstances.map {
```
longshorej: I feel like having two separate options (nrOfAgents, nrOfCores or similar) would be easier to use and more discoverable, but YMMV.

huntc: The arg reflects the arity of the arg in the CLI itself, so I think we're good here.

src/sbt-test/sbt-conductr/sandbox-conductr-roles/build.sbt
huntc: We were going to introduce new `conduct` CLI commands in order to display the member/agent info, so I don't think that the `sandbox ps` command itself needs updating: https://github.com/typesafehub/conductr-cli/issues/294

We could leverage that work once done, but I don't think it is essential to do here given that we're talking about a test which'll always be running on some flavour of unix.

PR #213, opened 2017-02-22
Using conductr-bundle-lib 1.6.1

markusjura:
Using conductr-bundle-lib 1.6.1 to support Lagom 1.3.x.


markusjura:
Tested this version with a modified `activator-lagom-java-chirper` Lagom app that uses Lagom 1.3.0-RC.

Issue #212, opened 2017-02-22
Set BundleKeys.conductrTargetVersion to ConductrVersion.V2


markusjura:
By default, the `BundleKeys.conductrTargetVersion` is set to `ConductrVersion.V1_1`.

ConductR 2.0.0 is released so we should change the default to `ConductrVersion.V2`.

As a consequence, Lagom bundles will generate the endpoints in the ACL format instead of the services format.

huntc:
Yeah, let's do it!

@longshorej Can you pick this up amongst your sbt-conductr tasks?

markusjura:
Sorry @longshorej. My day begins earlier so I've fixed this one already with PR https://github.com/typesafehub/sbt-conductr/pull/215.

PR #215, opened 2017-02-23
Use conductrTargetVersion V2_0 as default

markusjura:
ConductR 2.0.0 is released. Therefore, updating the default ConductR target version to `V2_0`.

Fixes https://github.com/typesafehub/sbt-conductr/issues/212#issuecomment-281841608.


markusjura:
Cherry picks:
- 2.2.x: https://github.com/typesafehub/sbt-conductr/pull/216

PR #218, opened 2017-02-24
Add support for newly added `sandbox run --no-default-features` argument

longshorej:
This adds support for newly added `sandbox run --no-default-features` argument. See https://github.com/typesafehub/conductr-cli/pull/325


longshorej:
Agreed on running tests with `--no-default-features` - I'll address that point in my other PR that is dealing with getting the tests to run.

PR #224, opened 2017-03-07
bindServices must be invoked exactly once.

ignasi35:
See http://www.lagomframework.com/documentation/1.3.x/java/ServiceImplementation.html: 

>  Lagom allows you to bind multiple services, however, bindServices() may only be invoked once, as this will bind a router for Play to use to route Lagom service calls, which if bound multiple times, will cause a Guice configuration error. The first binding you specify will be considered the primaryServiceBinding and its Descriptor’s name will be used to name your Service. This name will be used by Lagom as the default value to identify your microservice when interacting with other microservices.


markusjura:
Thanks for the contribution @ignasi35

PR #234, opened 2017-03-15
Fixes Lagom apps and removes unnecessary example

### LINKED ISSUE #233: scripted tests cover scenarios unsupported by Lagom
Several `sbt-scripted` tests use Lagom sample apps that are not compliant with Lagom's restrictions:

 * `bindServices` must be invoked only once [docs](http://www.lagomframework.com/documentation/1.3.x/java/ServiceImplementation.html#Implementing-services). E.g.  [lagom-bundle-plugin/multi-projects](https://github.com/typesafehub/sbt-conductr/blob/d5f42b30ca1c9390ad9a95039a9a87437791181a/src/sbt-test/lagom-bundle-plugin/multi-projects/payment-impl/src/main/java/impl/PaymentModule.java#L15-L16)
 * every service descriptor in a lagom system must use a unique name --> [multi-services-one-project-same-service-name](https://github.com/typesafehub/sbt-conductr/blob/d5f42b30ca1c9390ad9a95039a9a87437791181a/src/sbt-test/lagom-bundle-plugin/multi-services-one-project-same-service-name/test#L3-L3) can be removed.

ignasi35:
Fixes #233


ignasi35:
Removed `multi-services-one-project-same-service-name` since the only difference between that and `multiple-projects` is bundling several Service Descriptors that use the same name and that's not supported in Lagom.

PR #236, opened 2017-03-17
Split scripted tests into private and public

markusjura:
Splitting sbt scripted tests into private and public tests. The public tests are executed on Travis whereas the private tests are executed on our ci.lightbend.com Jenkins server.

The change is necessary to test the private tests with a Lightbend license.

To execute only private or public tests, two new directories under `sbt-test` have been created: private and public. Therefore Travis and Jenkins can run the corresponding tests by wildcard. The sbt scripted plugin unfortunately only supports two levels of sub directories under `sbt-test`. Therefore, the group directory and the test directory have now been merged, e.g. `bundle-plugin/basic` has been changed to `bundle-plugin-basic`: https://github.com/sbt/sbt/blob/0.13/scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala#L44


markusjura:
jenkins

markusjura:
jenkins

PR #238, opened 2017-03-17
Use max 3 ConductR instances (2.2.x)

markusjura:
The developer license we are using on our Jenkins server only supports 3 ConductR agents. This PR is adjusting the sbt tests so that we test with maximum 3 agents.


markusjura:
Cherry pick of https://github.com/typesafehub/sbt-conductr/pull/237 for branch 2.2.x

PR #256, opened 2017-04-19
Permit string path and re-write path combinations

huntc:
This commit permits the scenario where a string path and its rewrite path can be specified for all request methods. Without this fix, an sbt error would occur due to the enumeration "withName" method failing.


huntc:
https://github.com/typesafehub/sbt-conductr/pull/257 takes care of 2.2

PR #265, opened 2017-05-05
Use name bundle-status for status component (2.2.x)

markusjura:
Prior to this change, the name of the status component was derived from the bundle component. Example: A status component was named "my-bundle-status" if the bundle component name was "my-bundle". This logic makes it difficult to filter for the status component, especially if the bundle component contains the substring "status" as well.

Now, we name the status component always "bundle-status" to make it simplier to search for a status component, e.g. in the `conductr-cli` to auto-detect components.


markusjura:
Cherry pick of https://github.com/typesafehub/sbt-conductr/pull/263

PR #274, opened 2017-06-23
Update Lagom to 1.3.5 in tests

TimMoore:
This requires working around the conflicting `circuitBreakerMetricsProvider` definitions in `ConductRServiceLocatorComponents` and `LagomServiceClientComponents`.

See typesafehub/conductr-lib#145 for details.

### DIFF
--- a/sbt-conductr-tester/lagom-scala-bundle/build.sbt
+++ b/sbt-conductr-tester/lagom-scala-bundle/build.sbt
@@ -15,9 +15,6 @@ lazy val `play-service` = (project in file("play-service"))
   .enablePlugins(PlayScala, LagomPlay)
   .settings(
     routesGenerator := InjectedRoutesGenerator,
-    // Need to explicitly add the dependency to lagom-client 1.3.1 because 1.3.1-RC1 is being pulled in
-    libraryDependencies ++= Seq(
-      "com.lightbend.lagom" %% "lagom-scaladsl-client" % "1.3.1",
-      macwire)
+    libraryDependencies += macwire
   )
   .dependsOn(`lagom-service-api`)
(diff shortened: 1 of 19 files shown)


sbt-conductr-tester/lagom-scala-bundle/build.sbt
```diff
@@ -15,9 +15,6 @@ lazy val `play-service` = (project in file("play-service"))
   .enablePlugins(PlayScala, LagomPlay)
   .settings(
     routesGenerator := InjectedRoutesGenerator,
-    // Need to explicitly add the dependency to lagom-client 1.3.1 because 1.3.1-RC1 is being pulled in
```
TimMoore: This seems not to be required any longer. 1.3.5 is correctly evicting the RC.

In fact, I think this was true as of 1.3.1 (there were no RCs for 1.3.1, only 1.3.0). I'm guessing the comment was updated via an automatic search and replace.

TimMoore:
@huntc it was unrelated so I'll put it in a separate branch

huntc:
@TimMoore Separate commit will also work - if this PR fails then I can't merge it...

TimMoore:
That test also fails on my machine from the master branch, so I think it must be a difference between my laptop and everyone else's environments. I don't expect it to cause this PR to fail.

TimMoore:
Shows what I know 😦 https://travis-ci.org/typesafehub/sbt-conductr/builds/246072924#L1980-L2021

It's failing on master in Travis too, however, since https://travis-ci.org/typesafehub/sbt-conductr/builds/242759803

I have it passing locally now. I'll add the change to this branch.

PR #284, opened 2017-12-12
Upgrade to ConductR Lib 2.1.1

fsat:


fsat:
Only merge when ConductR Lib 2.1.1 is released.

PR #286, opened 2018-01-24
Lagom 1.4.0

renatocaval:


renatocaval:
@longshorej this is ready for review. 

I initially doubled the `lagom-bundle-plugin-play-and-lagom-service-scala` because the integration of play25/lagom13 is different than play26/lagom14, but later I realised that this is irrelevant for the test. 

The API differences have nothing to do with how to generate a bundle and deploy the apps.

WDYT?

(please, squash it)

PR #289, opened 2018-03-27
update conductr version

longshorej:


longshorej:
jenkins

longshorej:
jenkins

longshorej:
cc @JustinPihony this removes the token that we need to expire / may have already been expired

